PR #4, opened 2019-12-17
Remove Chains

devmotion:
It feels wrong to define and export `Chains` in this package even though the `Chains` struct is defined in MCMCChains. Since the `Chains` function definition in this package only throws an error, IMO it should be removed.


cpfiffer:
Good catch, thanks!

PR #6, opened 2020-01-12
Allow dispatch on chain type

cpfiffer:
Users can now pass in a keyword argument to `sample` to define the type of the chain they want to receive after sampling. It's the responsibility of package developers to support chain construction for different types. An example of usage might be

```julia
chain1 = sample(model, sampler, 1000; chain_type = NamedTuple)
chain2 = sample(model, sampler, 1000; chain_type = MCMCChains.Chains)
chain3 = sample(model, sampler, 1000; chain_type = Array)
```

If no keyword is specified, this defaults to whatever a package developer sets as the default constructor `bundle_samples`, or it simply returns the vector of `AbstractTransition` if the package developer has not created a catch-all overload of `bundle_samples`.

### DIFF
--- a/src/AbstractMCMC.jl
+++ b/src/AbstractMCMC.jl
@@ -198,6 +198,7 @@ function sample(
     s::SamplerType,
     N::Integer;
     progress::Bool=true,
+    chain_type::Type=Type{Any},
     kwargs...
 ) where {ModelType<:AbstractModel, SamplerType<:AbstractSampler}
     # Perform any necessary setup.
@@ -224,7 +225,11 @@ function sample(
     # Wrap up the sampler, if necessary.
     sample_end!(rng, ℓ, s, N, ts; kwargs...)
 
-    return bundle_samples(rng, ℓ, s, N, ts; kwargs...)
+    if chain_type == Type{Any}
+        return bundle_samples(rng, ℓ, s, N, ts; kwargs...)
+    else
+        return bundle_samples(rng, ℓ, s, N, ts, chain_type; kwargs...)
+    end
 end
 
 """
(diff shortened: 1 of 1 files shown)


src/AbstractMCMC.jl
```diff
@@ -224,7 +225,11 @@ function sample(
     # Wrap up the sampler, if necessary.
     sample_end!(rng, ℓ, s, N, ts; kwargs...)
 
-    return bundle_samples(rng, ℓ, s, N, ts; kwargs...)
+    if chain_type == Type{Any}
+        return bundle_samples(rng, ℓ, s, N, ts; kwargs...)
+    else
+        return bundle_samples(rng, ℓ, s, N, ts, chain_type; kwargs...)
+    end
```
devmotion: Maybe you could just call
```julia
return bundle_samples(rng, ℓ, s, N, ts, chain_type; kwargs...)
```
and define `bundle_samples` for the default chain type (`Type{Any}` or maybe a more specific custom singleton type). If `bundle_samples` without chain type is still needed (is it actually?) it could just be forwarded to the method with the default type. In general, I think it is easier for a user/developer to not have to deal with two different method signatures in case they want to implement some custom behaviour. At least that was my impression when implementing the `step!` function :wink: 

Another/additional idea would be to use `chain_type = nothing` as a default and call
```julia
return bundle_samples(rng, ℓ, s, N, ts, chain_type === nothing ? typeof(ts) : chain_type; kwargs...)
```
and then just define a default method
```julia
function bundle_samples(
    rng::AbstractRNG, 
    ℓ::AbstractModel, 
    s::AbstractSampler, 
    N::Integer, 
    ts::Vector{T},
    chain_type::T;
    kwargs...
) where {T<:AbstractTransition}
    return ts
end
```

devmotion:
In general, I think some of the static parameter types are not needed - e.g., `ModelType<:AbstractModel` is never defined or used (that also applies to some existing implementations).

cpfiffer:
Woops, I merged this before I saw your comments. I'll open a new one.

PR #8, opened 2020-01-12
Allow dispatch on chain type (take two)

cpfiffer:
I undid #6 because I missed some comments. I've addressed them here by pulling out some static type constraints and allowing for a single default `bundle_samples` method instead of two.

### DIFF
--- a/src/AbstractMCMC.jl
+++ b/src/AbstractMCMC.jl
@@ -180,6 +180,7 @@ function sample(
     s::AbstractSampler,
     N::Integer;
     progress::Bool=true,
+    chain_type::Type=Any,
     kwargs...
 )
     # Perform any necessary setup.
@@ -206,7 +207,7 @@ function sample(
     # Wrap up the sampler, if necessary.
     sample_end!(rng, ℓ, s, N, ts; kwargs...)
 
-    return bundle_samples(rng, ℓ, s, N, ts; kwargs...)
+    return bundle_samples(rng, ℓ, s, N, ts, chain_type; kwargs...)
 end
 
 """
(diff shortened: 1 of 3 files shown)


src/AbstractMCMC.jl
```diff
@@ -116,18 +116,18 @@ DefaultCallback(N::Int) = DefaultCallback(ProgressMeter.Progress(N, 1))
 
 function init_callback(
     rng::AbstractRNG,
-    ℓ::ModelType,
+    ℓ::AbstractModel,
     s::SamplerType,
     N::Integer;
     kwargs...
-) where {ModelType<:AbstractModel, SamplerType<:AbstractSampler}
+) where {SamplerType<:AbstractSampler}
```
devmotion: I guess `SamplerType` is not needed either? This comment also applies to most of the other function definitions below.

src/AbstractMCMC.jl
```diff
@@ -224,13 +225,29 @@ function sample(
...
+    return bundle_samples(rng, ℓ, s, N, ts, chain_type; kwargs...)
+end
+
+function bundle_samples(
+    rng::AbstractRNG, 
+    ℓ::AbstractModel, 
+    s::SamplerType, 
+    N::Integer, 
+    ts::Vector{T},
+    chain_type::Type = Type{Any};
```
devmotion: I'm still wondering if there's a reason for allowing to call `bundle_samples` without `chain_type`. As I mentioned in the previous PR, IMO it's easier to work with consistent method signatures - that still allows to specify a default method with `chain_type::Type{Any}` that can be changed for specific models, samplers, and transitions.

src/AbstractMCMC.jl
```diff
@@ -224,13 +225,29 @@ function sample(
...
+
+function bundle_samples(
+    rng::AbstractRNG, 
+    ℓ::AbstractModel, 
+    s::SamplerType, 
+    N::Integer, 
+    ts::Vector{T},
+    chain_type::Type = Type{Any}; 
+    kwargs...
+) where {SamplerType<:AbstractSampler, T<:AbstractTransition}
```
devmotion: It seems `T<:AbstractTransition` is not needed either, one could just write `ts::Vector{<:AbstractTransition}`?

src/AbstractMCMC.jl
cpfiffer: This makes it completely opt-in without much cost. If you want to change your `bundle_samples` code to work with different outputs, great! Otherwise, this is just a nice default that's automatically called for you. I don't really care for requiring people to have some kind of formalized output.

cpfiffer:
Alright, I made it so the `bundle_samples` signature is consistent, and added a test to show different chain types.

devmotion:
Looks good to me.

cpfiffer:
Thanks, @devmotion!

PR #9, opened 2020-01-24
Simplify psample

devmotion:
This PR updates `psample` and adds some tests for the interface and the default implementation.

### DIFF
--- a/test/interface.jl
+++ b/test/interface.jl
@@ -1,0 +1,64 @@
+struct MyModel <: AbstractModel end
+
+struct MyTransition <: AbstractTransition
+    a::Float64
+    b::Float64
+end
+
+struct MySampler <: AbstractSampler end
+
+AbstractMCMC.transition_type(::MySampler) = MyTransition
+
+struct MyChain <: AbstractChains
+    as::Vector{Float64}
+    bs::Vector{Float64}
+end
+
+function AbstractMCMC.step!(
+    rng::AbstractRNG,
+    model::MyModel,
+    sampler::MySampler,
+    N::Integer;
+    kwargs...
+)
+    a = rand(rng)
+    b = randn(rng)
+
+    return MyTransition(a, b)
+end
+
+function AbstractMCMC.step!(
+    rng::AbstractRNG,
+    model::MyModel,
+    sampler::MySampler,
+    N::Integer,
+    transition::Union{Nothing,MyTransition};
+    kwargs...
+)
+    a = rand(rng)
+    b = randn(rng)
+
+    return MyTransition(a, b)
+end
+
+function AbstractMCMC.bundle_samples(
+    rng::AbstractRNG,
+    model::MyModel,
+    sampler::MySampler,
+    N::Integer,
+    transitions::Vector{MyTransition};
+    kwargs...
+)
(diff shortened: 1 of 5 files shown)


test/interface.jl
```diff
@@ -0,0 +1,64 @@
...
+    bs::Vector{Float64}
+end
+
+function AbstractMCMC.step!(
+    rng::AbstractRNG,
+    model::MyModel,
+    sampler::MySampler,
+    N::Integer;
+    kwargs...
+)
```
devmotion: It would be nice not having to define this special method for the initial implementation. It could be avoided if one would define
```julia
step!(rng, model, sampler, N = 1; kwargs) = step!(rng, model, sampler, N, nothing; kwargs...)
```
as the default implementation instead of the other way around, as it is currently done in AbstractMCMC. One could still use the shorthand call without `nothing` but would just have to overload the methods with `transition` in custom implementations such as in Turing. The advantage would be that in cases where the first step is equal to all other steps one would just implement one method `step!(...., transition::Union{Nothing,Transition}; ...)` (such as in the case of the Gibbs sampler, https://github.com/TuringLang/Turing.jl/blob/867752a2da1ec0d8e2fe9737a3fe4bf0cb19adcd/src/inference/gibbs.jl#L130-L213), and in the other cases one would just implement two methods, as before, with type signature `step!(..., ::Nothing; kwargs...)` and `step!(..., ::Transition; kwargs...)`. What do you think?

test/interface.jl
```diff
@@ -0,0 +1,64 @@
+struct MyModel <: AbstractModel end
+
+struct MyTransition <: AbstractTransition
+    a::Float64
+    b::Float64
+end
+
+struct MySampler <: AbstractSampler end
+
+AbstractMCMC.transition_type(::MySampler) = MyTransition
```
devmotion: Where is `transition_type` actually used or needed? I looked around in AbstractMCMC and Turing, and it seems it is only called in the default implementation of `sample(...)` in AbstractMCMC. Is that correct?

Because if that would be the case, it could be dropped quite easily. One could just sample the first step (which is handled differently inside of the for loop anyway), and then create a vector based on the returned type, which is then filled with the samples from all remaining steps. In the case `N < 1` one could throw an error, so one doesn't have to bother with type stability issues.

test/interface.jl
cpfiffer: I'm fine with that -- I don't particularly care for the current approach where there's two distinct signatures, and I think this is a good solution.

test/interface.jl
cpfiffer: My initial rationalization was that maybe someone would want to overload either method -- if someone didn't much care for the default version of `transitions_init` and wanted to generate some weird struct to store samples in, they could do so without limiting it to the vector form. `transition_type` is just an ease of use thing to allow someone to use the default functionality fairly quickly. With all the interface stuff I really just wanted to provide defaults that do everything for you when you don't want to, but you can overload if you care a great deal.

We probably could instead just draw the first sample without too much trouble and use that as the type -- I've got no issues with that.

cpfiffer:
I'll merge this for now. Your other comments are probably better as other PRs.

PR #12, opened 2020-01-27
Change handling of transitions

devmotion:
I've thought a bit more about `transition_type`, and how to improve on what I mentioned in https://github.com/TuringLang/AbstractMCMC.jl/pull/9#discussion_r370753197.

In this PR I propose the following changes:
- Remove `transition_type` and instead just obtain the first transition
- Pass the initial transition to `transitions_init`, whose default implementation just creates a vector for `N` transitions of the type of the initial transition
- Add a function `transitions_save!` to which (among others) the transitions container, the current iteration, and the current transition is passed; the default implementation for vector-type containers just saves the current transition in the container
- Remove `AbstractTransition`
- Do not export anything

The main advantages with this approach are:
- Users/developers do not have to define `transition_type` - as long as their `step!` implementation is type stable it just works
- It seems trivial to, e.g., add a wrapper sampler
  ```julia
  struct StreamingSampler{S<:AbstractSampler,I}
      io::I
      sampler::S
  end

  stream(io, sampler::AbstractSampler) = StreamingSampler(io, sampler)

  AbstractMCMC.sample_init!(rng::AbstractRNG, model::AbstractModel, sampler::StreamingSampler, N::Integer; kwargs...) = sample_init!(rng, model, sampler.sampler, N; kwargs...)
  AbstractMCMC.sample_end!(rng::AbstractRNG, model::AbstractModel, sampler::StreamingSampler, N::Integer, transitions; kwargs...) = sample_end!(rng, model, sampler.sampler, N, transitions; kwargs...)
  AbstractMCMC.bundle_samples!(rng::AbstractRNG, model::AbstractModel, sampler::StreamingSampler, N::Integer, ::Nothing; kwargs...) = nothing

  AbstractMCMC.transitions_init!(transition, ::AbstractModel, sampler::StreamingSampler, ::Integer; kwargs...) = nothing
  AbstractMCMC.transitions_save!(::Nothing, ::Integer, transition, ::AbstractModel, ::StreamingSampler, ::Integer) = write(sampler.io, transition)
  ```
  as suggested in https://github.com/TuringLang/Turing.jl/issues/1053
- One does not have to wrap parameter vectors or named tuples inside of a custom `AbstractTransition` subtype anymore - if `step!` just returns a parameter vector or a named tuple, the default implementation of `sample` will just yield a vector of parameter vectors or named tuples
- It does not seem useful to pollute the namespace by exporting methods such as `sample_init!` or `sample_end!` (which many people do not want to call directly and have to be imported anyway if implemented for specific types). In general, I think in every concrete implementation of AbstractModel/AbstractSampler one just has to decide which methods of AbstractMCMC to export (probably just `sample` and `psample`).

### DIFF
--- a/src/AbstractMCMC.jl
+++ b/src/AbstractMCMC.jl
@@ -57,28 +57,6 @@ An `AbstractModel` represents a generic model type that can be used to perform i
 abstract type AbstractModel end
 
 """
-    AbstractTransition
-
-The `AbstractTransition` type describes the results of a single step
-of a given sampler. As an example, one implementation of an
-`AbstractTransition` might include be a vector of parameters sampled from
-a prior distribution.
-
-Transition types should store a single draw from any sampler, since the
-interface will sample `N` times, and store the results of each step in an
-array of type `Array{Transition<:AbstractTransition, 1}`. If you were
-using a sampler that returned a `NamedTuple` after each step, your
-implementation might look like:
-
-```
-struct MyTransition <: AbstractTransition
-    draw :: NamedTuple
-end
-```
-"""
-abstract type AbstractTransition end
-
-"""
     AbstractCallback
 
 An `AbstractCallback` types is a supertype to be inherited from if you want to use custom callback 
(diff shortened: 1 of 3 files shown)


cpfiffer:
I'm a fan! I'll sit down with this sometime this evening and give some more thoughtful comments if I can find anything.

devmotion:
Should we merge it? :smiley:

cpfiffer:
Yeah, sorry. Got caught up in some stuff and it slipped my mind.

PR #13, opened 2020-02-02
Add iterator interface

cpfiffer:
Allows the use of `steps!`, which returns an iterator that continuously draws samples. Currently it's pretty bare bones, but I'm open to suggestions on what other functionality that could be included.

Usage:

```julia
# Sample until 1000 iterations are hit.
for (total, transition) in enumerate(AbstractMCMC.steps!(MyModel(), MySampler()))
    println(transition)
    
    if total >= 1000
        break
    end
end
```

### DIFF
--- a/src/AbstractMCMC.jl
+++ b/src/AbstractMCMC.jl
@@ -415,8 +415,8 @@ end
 ##################
 # Iterator tools #
 ##################
-struct Stepper{ModelType<:AbstractModel, SamplerType<:AbstractSampler, K}
-    rng::AbstractRNG
+struct Stepper{A<:AbstractRNG, ModelType<:AbstractModel, SamplerType<:AbstractSampler, K}
+    rng::A
     model::ModelType
     s::SamplerType
     kwargs::K
@@ -460,8 +460,7 @@ function steps!(
     kwargs...
 )
     sample_init!(rng, model, s, 0)
-    iterable = Stepper(rng, model, s, kwargs)
+    return Stepper(rng, model, s, kwargs)
 end
-
 
 end # module AbstractMCMC
--- a/src/AbstractMCMC.jl
+++ b/src/AbstractMCMC.jl
@@ -415,23 +415,20 @@ end
 ##################
 # Iterator tools #
 ##################
-struct Stepper{ModelType<:AbstractModel, SamplerType<:AbstractSampler, T, K}
+struct Stepper{ModelType<:AbstractModel, SamplerType<:AbstractSampler, K}
     rng::AbstractRNG
     model::ModelType
     s::SamplerType
-    t::T
     kwargs::K
 end
 
-function Base.iterate(stp::Stepper, state=(stp.t, 0))
-    t, i = state
-    new_t = step!(stp.rng, stp.model, stp.s, 1, t; stp.kwargs...)
-    
-    return t, (new_t, i+1)
+function Base.iterate(stp::Stepper, state=nothing)
+    t = step!(stp.rng, stp.model, stp.s, 1, state; stp.kwargs...)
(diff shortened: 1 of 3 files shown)


devmotion:
I guess one should also implement
```julia
Base.IteratorSize(::Type{<:Stepper}) = Base.IsInfinite()
Base.eltype(::Type{<:Stepper{<:AbstractModel,<:AbstractSampler,T}}) where T = T
```
(see https://docs.julialang.org/en/v1/base/collections/). I guess, in particular `eltype` might be needed for strictly typed arrays when performing something like `collect(Iterators.take(steps!(model, sampler), 20))`.

cpfiffer:
Updated. I missed those when I was implementing this originally. Most of this was random code I wrote many months ago and thought should be kept around.

Are there any convenience functions we should build here? I don't really know what anyone will use this for.

devmotion:
Maybe add a test for these methods as well?

Otherwise, I don't really know what else to add :man_shrugging: I guess, one can add any convenience functions at a later point as well, if someone comes up with a good idea.

src/AbstractMCMC.jl
```diff
@@ -411,4 +411,60 @@ function psample(
...
 
+
+##################
+# Iterator tools #
+##################
+struct Stepper{ModelType<:AbstractModel, SamplerType<:AbstractSampler, T, K}
+    rng::AbstractRNG
+    model::ModelType
+    s::SamplerType
+    t::T
```
devmotion: Just noticed that it feels slightly inconsistent to add the initial transition to the struct but make all future transitions part of the state. I guess, the natural approach would be to remove `t::T` and define the initial state as `nothing` (I don't really see why you would need the iteration index?), i.e.,
```julia
function Base.iterate(stp::Stepper, state=nothing)
    t = step!(stp.rng, stp.model, stp.s, 1, state; stp.kwargs...)
    return t, t
end
```

cpfiffer: What's a convenient way to track `eltype` in this case? Storing the type explicitly in the `Stepper` struct?

devmotion: In that case one should probably not implement `eltype` at all but only
```julia
Base.IteratorEltype(::Type{<:Stepper}) = Base.EltypeUnknown()
```
as [mentioned in the docs](https://docs.julialang.org/en/v1/base/collections/#Base.IteratorEltype). Then `collect` should still return a concretely typed array (I assume the output type is determined by the type of the first iterate).

src/AbstractMCMC.jl
```diff
@@ -411,4 +411,57 @@ function psample(
     return reduce(chainscat, chains)
 end
 
+
+##################
+# Iterator tools #
+##################
+struct Stepper{ModelType<:AbstractModel, SamplerType<:AbstractSampler, K}
+    rng::AbstractRNG
```
devmotion: I suggest to replace this with a concrete type by adding a type parameter.

src/AbstractMCMC.jl
```diff
@@ -411,4 +411,57 @@ function psample(
...
+end
+
+function steps!(
+    rng::AbstractRNG,
+    model::AbstractModel,
+    s::AbstractSampler,
+    kwargs...
+)
+    sample_init!(rng, model, s, 0)
+    iterable = Stepper(rng, model, s, kwargs)
```
devmotion: The assignment could be removed.

devmotion:
What's holding this back? I guess, if my questions/suggestions are addressed, this could be merged?

cpfiffer:
Only thing holding this back is my schedule, but I believe I've addressed your comments. We can merge after tests pass.

PR #17, opened 2020-02-26
Use ProgressLogging instead of ProgressMeter

devmotion:
This PR removes `AbstractCallback` and instead allows the user to provide a custom callback function. Moreover, progress logging is separated from callback handling and uses ProgressLogging instead of ProgressMeter.

ProgressLogging uses Julia's logging system to create, update, and display progress bars. It integrates nicely with Juno and can also be used in the terminal with TerminalLoggers. Among others, it is used by JuliaDiffEq and Flux for displaying progress bars. The current implementation does not make use of the updated functionality (e.g., nested progress bars) and API in ProgressLogging 0.1.1 since it is not fully supported yet by, e.g., TerminalLoggers.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,9 +1,14 @@
 using AbstractMCMC
 using AbstractMCMC: sample, psample, steps!
+using TerminalLoggers: TerminalLogger
 
+using Logging: global_logger
 using Random
 using Statistics
 using Test
+
+# install progress logger
+global_logger(TerminalLogger(right_justify=120))
 
 include("interface.jl")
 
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -51,13 +56,21 @@ include("interface.jl")
...
-        chain2 = sample(MyModel(), MySampler(), 100; progress = true, sleepy = true, chain_type = MyChain)
+        chain1 = sample(MyModel(), MySampler(), 100; sleepy = true)
+        chain2 = sample(MyModel(), MySampler(), 100; sleepy = true, chain_type = MyChain)
 
         @test chain1 isa Vector{MyTransition}
         @test chain2 isa MyChain
     end
 
+    @testset "Suppress output" begin
+        sample(MyModel(), MySampler(), 100; progress = false, sleepy = true)
```
devmotion: I'm not sure how to properly test that nothing is displayed here.

cpfiffer: Maybe we could pass in an `IO` variable, write the output to a file, and check that the file's length is zero? Although I can't see how you would pass an IO context to ProgressLogging.

devmotion: I found `collect_test_logs` in the Test standard library which seems to be useful.

cpfiffer:
I like it. Do you want to go up a minor version for this?

devmotion:
To 0.5, you mean?

cpfiffer:
Yeah.

Issue #18, opened 2020-03-02
ProgressLogging default environments


cpfiffer:
I just noticed that the new ProgressLogging.jl backend does not actually have any display functionality by default, and that downstream users do not get any automatic progress bar updates. 

Is there a way to detect which environment someone is in (either Juno or a terminal) and to use the appropriate progress viewer automatically? I wasn't able to find a good way to do so.

devmotion:
Actually, IMO detecting the environment and using a specific logging viewer is against the purpose of ProgressLogging. The purpose of ProgressLogging is to provide an easy, backend-agnostic way of defining progress logging with the help of the Julia logging system. Users then can decide if they, e.g., want to drop the logging messages, archive them to files, or display them in the terminal.

That being said, Juno will display a native progress bar automatically. For displaying progress bars in the REPL, users can install an appropriate logger such as `TerminalLogger` globally
```julia
using TerminalLoggers: TerminalLogger
using Logging: global_logger

global_logger(TerminalLogger())
```
or only locally when calling `sample`
```julia
using TerminalLoggers: TerminalLogger
using Logging: with_logger

with_logger(TerminalLogger()) do
	sample(....)
end
```
If `TerminalLogger` should be enabled by default, it [can be loaded in the startup file `~/.julia/config/startup.jl`](https://c42f.github.io/TerminalLoggers.jl/stable/#Installation-and-setup-1).

In principle, this also works with Jupyter notebooks, however, the output of `TerminalLogger` seems to fill up the the notebook since it is not cleared. As a workaround, one can use [ConsoleProgressMonitor.jl](https://github.com/tkf/ConsoleProgressMonitor.jl) which uses ProgressMeter for printing progress bars (but is superseded by TerminalLoggers). Again, ConsoloeProgressMonitor could be loaded globally for every Jupyter notebook by modifying the startup file `~/.julia/config/startup_ijulia.jl`.

devmotion:
See also https://github.com/c42f/TerminalLoggers.jl/issues/25

cpfiffer:
One of the big reasons we provided `ProgressMeter` in the first place was to show sampling as it occurred by default. As much as I like hands-off options, Turing (and others) have progress bars by default because many people would rather Turing "just worked" rather than specify their own terminal logger. I think we should consider enforcing a default so that there is literally any kind of visual representation of progress, rather than just a blank screen for seconds or minutes.

cpfiffer:
We could have something like 

[...]

PR #20, opened 2020-03-04
Display warning message if no known progress logging frontend is loaded

devmotion:
This PR is a take on https://github.com/TuringLang/AbstractMCMC.jl/issues/18#issuecomment-594759943. I'm not sure yet how reliable this approach detects if one of the known and recommend progress logging frontends is set up, or not.

### DIFF
--- a/src/AbstractMCMC.jl
+++ b/src/AbstractMCMC.jl
@@ -1,6 +1,7 @@
 module AbstractMCMC
 
 import ProgressLogging
+import Requires
 import StatsBase
 using StatsBase: sample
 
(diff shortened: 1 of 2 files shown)


devmotion:
Not completely sure how to test this though, maybe one has to use SafeTestsets or manually create some modules?

src/AbstractMCMC.jl
```diff
@@ -77,6 +92,9 @@ function StatsBase.sample(
     # Perform any necessary setup.
     sample_init!(rng, model, sampler, N; kwargs...)
 
+    # Check progress logger frontends.
+    progress && checkprogresslogger()
```
cpfiffer: This doesn't seem like `checkprogresslogger()` will be run unless `progress==false`. Should this be `!progress && checkprogresslogger()`?

cpfiffer:
> Not completely sure how to test this though, maybe one has to use SafeTestsets or manually create some modules?

Maybe

```julia
@safetestset "Warning test" begin 
    @test_logs (:warn, warning_message) using AbstractMCMC 
end
```

Although I'm not sure if this will scope correctly.

src/AbstractMCMC.jl
devmotion: It performs the check if `progress` logging is enabled - that seems reasonable, doesn't it? If progress logging is disabled we do not care about the loaded packages, I guess.

cpfiffer: Woops. Misread `&` as `|` somehow.

PR #22, opened 2020-03-06
Fix module of progress logger

devmotion:
When I updated Turing VI to use `@ifwithprogresslogger` as well I noticed a stupid bug which limits the reuse of the macro in other packages (currently it only works correctly when used inside of AbstractMCMC).


devmotion:
@cpfiffer Good to merge and release?

PR #27, opened 2020-03-27
Install CompatHelper as GitHub Action

sethaxen:


.github/workflows/CompatHelper.yml
```diff
@@ -0,0 +1,19 @@
+name: CompatHelper
+
+on:
+  schedule:
+    - cron: '00 00 * * *'
```
devmotion: Can we change the frequency to once a day? IMO it's not necessary to run it every hour.

.github/workflows/CompatHelper.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  CompatHelper:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: 1.3
```
devmotion: Maybe use version 1.4?
```suggestion
          version: 1.4
```

.github/workflows/CompatHelper.yml
devmotion: Oh sorry, it seems it's already once a day... I'm always confused by the cron notation :smile:

sethaxen: No worries, yeah I just used the boilerplate from CompatHelper's README. Been a while since I touched cron.

PR #32, opened 2020-03-27
Simplify callback interface

devmotion:
This PR is a proposal for simplifying the callback interface.

The removed arguments are known before sampling and hence, if needed, the callback can be defined as closure over these arguments. The RNG, the model, and the sampler are kept since they are not known (or rather their copies have not been created yet) before parallel sampling. Alternatively, these arguments could be included only in the callback interface for parallel sampling.


cpfiffer:
I like that the callbacks are consistent between parallel and serial sampling, so let's leave it that way. Thanks!

PR #33, opened 2020-04-01
Add multicore sampling

devmotion:
This PR adds multicore sampling as an alternative to the current parallelization using multithreading.

### DIFF
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -1,4 +1,4 @@
-# Default implementations of `sample` and `psample`.
+# Default implementations of `sample`.
 
 function StatsBase.sample(
     model::AbstractModel,
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -189,7 +189,7 @@ function mcmcsample(
     N::Integer,
     nchains::Integer;
     progress = true,
-    progressname = "Parallel sampling",
+    progressname = "Parallel sampling ($(Threads.nthreads()) threads)",
     kwargs...
 )
     # Copy the random number generator, model, and sample for each thread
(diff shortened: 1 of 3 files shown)


src/sample.jl
```diff
@@ -19,25 +19,28 @@ function StatsBase.sample(
     return mcmcsample(rng, model, sampler, arg; kwargs...)
 end
 
-function psample(
+function StatsBase.sample(
```
cpfiffer, line 22: If you want this to go into `master`, there needs to be something (possibly with a dep warning) 

```julia
psample(rng, model, spl, N, N_chains) = sample(rng, model, spl, ParallelThreads(), N, N_chains)
```

so we don't break any existing parallel code that relied on the previous default of thread parallelism.

cpfiffer: Nevermind, just saw these down below.

cpfiffer:
I think we should throw a warning for both cases if the number of threads/workers is one -- in this case, the user is not going to get any parallelism and should be notified.

devmotion:
I've added a warning in case the chains are not actually sampled in parallel. Moreover, I renamed the singletons to `MCMCThreads` and `MCMCDistributed`, somehow `ParallelThreads` and `ParallelDistributed` felt a bit too general.
At least for the tests it felt useful to export these singletons and `sample` but I can revert this if you thing AbstractMCMC shouldn't export anything.

cpfiffer:
I think they're fine to export. Looks good to me!

cpfiffer:
Semver suggests that deprecation merit a minor release, but we're not using full semver right -- I'm going to bump us up to 1.0.0 for the moment. We'll get all the stuff in the `dev` branch in 2.0.

mileslucas:
Any idea of performance or gotchas on multi core vs multi thread?

devmotion:
I guess you just have to benchmark which one is best for your particular use case :woman_shrugging: 

The most detailed information about multi-core vs multi-threading in Julia in general you can probably find in the [Julia documentation](https://docs.julialang.org/en/v1/manual/parallel-computing/#Parallel-Computing-1).

Issue #31, opened 2020-03-27
[RFC] Sampler states


devmotion:
I think the current approach of dealing with states of samplers, e.g., in Turing is flawed. Currently, one defines an initial state when initializing the sampler. This is problematic since often the concrete types of the states are not known during initialization which in turn then requires setting up dummy states and states with abstractly typed fields which potentially impacts performance.

With the current API of AbstractMCMC it is possible to avoid this issue by making the sampler state a part of the transition generated by `step!` and then implementing `transitions_init` and `transitions_save!` such that only the part of the transitions without the sampler state is actually saved (as done for the Gibbs sampler in Turing and proposed for the DynamicNUTS sampler in https://github.com/TuringLang/Turing.jl/pull/1186). Moreover, since only the samples + statistics are stored, it is impossible to resume sampling since the state of the sampler is not known anymore.

So as mentioned in https://github.com/TuringLang/Turing.jl/pull/1186, one could switch to a setup with (more or less) stateless samplers and explicit states, something along the lines of:

```julia
function mcmcsample(
    rng::Random.AbstractRNG,
    model::AbstractModel,
    sampler::AbstractSampler,
	N::Integer;
    progress = true,
    progressname = "Sampling",
    callback = (args...; kwargs...) -> nothing,
    chain_type::Type=Any,
    kwargs...
)
    # Check the number of requested samples.
    N > 0 || error("the number of samples must be ≥ 1")

    @ifwithprogresslogger progress name=progressname begin
        # Obtain the initial state.
        state = initialstate(rng, model, sampler, N; kwargs...)

        # Run callback.
        callback(rng, model, sampler, N, 1, state; kwargs...)

        # Save the initial sample.
        samples = initsamples(state, model, sampler, N; kwargs...)
        savesample!(samples, 1, state, model, sampler, N; kwargs...)

        # Update the progress bar.
        progress && ProgressLogging.@logprogress 1/N

        # Step through the sampler.
        for i in 2:N
            # Obtain the updated state.
            state = step(rng, model, sampler, N, state; kwargs...)

            # Run callback.
            callback(rng, model, sampler, N, i, state; kwargs...)

            # Save the sample.
            savesample!(samples, i, state, model, sampler, N; kwargs...)

            # Update the progress bar.
            progress && ProgressLogging.@logprogress i/N
        end
    end

    # Compute the resulting MCMC chain.
[...]

PR #35, opened 2020-05-06
Use BangBang to widen containers if needed

devmotion:
This PR addresses the problem with `transition_save!` for stochastic control flow mentioned in https://github.com/TuringLang/DynamicPPL.jl/pull/105#issue-413520535. It changes the default implementations for writing and appending to a container of transitions from `setindex!` and `push!` to `BangBang.setindex!!` and `BangBang.push!!` which widen the container or return a new immutable instance if needed. For consistency with these default implementations and to indicate that one has to return the (potentially new) container, I renamed `transitions_save!` to ``save!!` (and `transitions_init` to just `transitions`).

I'm wondering, should we start with merging such breaking changes to a dev branch first? I assume that, e.g., solving or addressing https://github.com/TuringLang/AbstractMCMC.jl/issues/31 will also require some changes of the interface, and hence maybe it would make sense to gather multiple breaking changes before merging them into master and releasing them?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,9 @@ branches:
 os:
   - linux
   - osx
-julia:
-  - 1.0
 matrix:
   include:
+    - julia: 1.0
     - julia: 1
       env: JULIA_NUM_THREADS=1
     - julia: 1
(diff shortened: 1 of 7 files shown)


cpfiffer:
I think we're fine to go into `master` in this repo, since AbstractMCMC doesn't change much and isn't very prone to bugs.

mohamed82008:
Is this good to go?

PR #39, opened 2020-05-18
Add a `Sample` transducer

devmotion:
I don't know if this PR should ever be merged, it's mainly me playing around with and learning about transducers. I assume transducers could be useful in AbstractMCMC since it's easy to combine and nest them with good performance (see, e.g., https://github.com/JuliaLang/julia/pull/33526). Some rudimentary progress logging can be achieved with `Transducers.withprogress`. Parallel reductions (multicore or multithreaded) can be achieved easily, and it's also easy to only save every nth element, drop samples, or iterate until some criterion is met.

Currently the implementation feels a bit weird since the transducer is mutated, ideally stepping would not mutate the model or the sampler (not sure how to deal with the RNG) (see https://github.com/TuringLang/AbstractMCMC.jl/issues/31).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -15,6 +15,7 @@ ProgressLogging = "33c8b6b6-d38a-422a-b730-caa89a2f386c"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 TerminalLoggers = "5d786b92-1e48-4d6f-9151-6b4477ca9bed"
+Transducers = "28d57a85-8fef-5791-bfe6-a80928e7c999"
 
 [compat]
 BangBang = "0.3.19"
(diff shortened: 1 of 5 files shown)


cpfiffer:
Neat! I've always wanted to see some transducers stuff in here. I'm not opposed to merging it in.

As to the RNG, I wonder if you could generate a seed at each step, pass it forward, and use the seed to make a new RNG on a per-step basis. Though maybe this is costly to do. Either way I'd like to see these be fully immutable/stateful.

devmotion:
> As to the RNG, I wonder if you could generate a seed at each step, pass it forward, and use the seed to make a new RNG on a per-step basis. Though maybe this is costly to do. Either way I'd like to see these be fully immutable/stateful.

I guess one possibility would be to make the RNG part of the private state of the transducer, then you don't have to create a RNG in every step. The question still remains what to do in the initial step. One could save a seed and an RNG as template, and then initialize a copy of the RNG with the seed in the initial step. That should then give the exact same samples every time - but I'm not sure if we actually want that and if the initialization step is too costly.

devmotion:
Yeah, I guess we can just let it evolve and grow and not export or use it properly right now.

PR #40, opened 2020-05-19
Fix undefined reference error when type widening

mohamed82008:
While testing stochastic control flow in Turing and DynamicPPL, I ran into an issue where `BangBang.setindex!!` tries to copy all of the old vector of transitions. Some of the elements of this vector would be undefined so it would throw and undefined reference error. This PR uses `BangBang.push!!` and `sizehint!` instead.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -151,7 +151,6 @@ function save!!(
     ::AbstractSampler;
     kwargs...
 )
-    @assert length(transitions) == iteration - 1
     return BangBang.push!!(transitions, transition)
 end
 
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -125,12 +125,11 @@ Save the `transition` of the MCMC `sampler` at the current `iteration` in the co
 `transitions`.
 
 The function can be called with and without a predefined size `N`. By default, AbstractMCMC
-uses ``setindex!`` and ``push!!`` from the Julia package
-[BangBang](https://github.com/tkf/BangBang.jl) to write to and append to the container,
-and widen the container type if needed.
+uses ``push!!`` from the Julia package [BangBang](https://github.com/tkf/BangBang.jl) to
+append to the container, and widen its type if needed.
 """
 function save!!(
-    transitions,
+    transitions::Vector,
     transition,
     iteration::Integer,
     ::AbstractModel,
@@ -139,7 +138,7 @@ function save!!(
     kwargs...
 )
     new_ts = BangBang.push!!(transitions, transition)
-    typeof(new_ts) !== typeof(transitions) && Base.sizehint!(new_ts, N)
+    new_ts !== transitions && sizehint!(new_ts, N)
     return new_ts
 end
 
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -104,7 +104,7 @@ function transitions(
     kwargs...
 )
     ts = Vector{typeof(transition)}(undef, 0)
-    Base.sizehint!(ts, N)
+    sizehint!(ts, N)
     return ts
 end
 
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -103,7 +103,9 @@ function transitions(
     N::Integer;
     kwargs...
 )
-    return Vector{typeof(transition)}(undef, N)
+    ts = Vector{typeof(transition)}(undef, 0)
+    Base.sizehint!(ts, N)
+    return ts
 end
 
 function transitions(
@@ -112,7 +114,7 @@ function transitions(
     ::AbstractSampler;
     kwargs...
 )
-    return Vector{typeof(transition)}(undef, 1)
+    return Vector{typeof(transition)}(undef, 0)
 end
 
 """
@@ -133,10 +135,13 @@ function save!!(
     iteration::Integer,
     ::AbstractModel,
     ::AbstractSampler,
-    ::Integer;
+    N::Integer;
     kwargs...
 )
-    return BangBang.setindex!!(transitions, transition, iteration)
+    @assert length(transitions) == iteration - 1
+    new_ts = BangBang.push!!(transitions, transition)
+    typeof(new_ts) !== typeof(transitions) && Base.sizehint!(new_ts, N)
+    return new_ts
 end
 
 function save!!(
@@ -147,6 +152,7 @@ function save!!(
     ::AbstractSampler;
     kwargs...
 )
+    @assert length(transitions) == iteration - 1
     return BangBang.push!!(transitions, transition)
 end
 
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -138,7 +138,6 @@ function save!!(
     N::Integer;
     kwargs...
 )
-    @assert length(transitions) == iteration - 1
     new_ts = BangBang.push!!(transitions, transition)
     typeof(new_ts) !== typeof(transitions) && Base.sizehint!(new_ts, N)
     return new_ts
@@ -160,4 +159,3 @@ Base.@deprecate transitions_init(transition, model::AbstractModel, sampler::Abst
 Base.@deprecate transitions_init(transition, model::AbstractModel, sampler::AbstractSampler; kwargs...) transitions(transition, model, sampler; kwargs...) false
 Base.@deprecate transitions_save!(transitions, iteration::Integer, transition, model::AbstractModel, sampler::AbstractSampler; kwargs...) save!!(transitions, transition, iteration, model, sampler; kwargs...) false
 Base.@deprecate transitions_save!(transitions, iteration::Integer, transition, model::AbstractModel, sampler::AbstractSampler, N::Integer; kwargs...) save!!(transitions, transition, iteration, model, sampler, N; kwargs...) false
-


src/sample.jl
```diff
@@ -81,8 +81,8 @@ function mcmcsample(
         callback(rng, model, sampler, transition, 1)
 
         # Save the transition.
-        transitions = AbstractMCMC.transitions(transition, model, sampler, N; kwargs...)
-        transitions = save!!(transitions, transition, 1, model, sampler, N; kwargs...)
+        transitions = [transition]
```
devmotion, line 84: We don't want to do that, the interface `transitions` is defined so that downstream packages can use other containers than vectors.

src/sample.jl
```diff
@@ -81,8 +81,8 @@ function mcmcsample(
...
+        Base.sizehint!(transitions, N)
```
devmotion, line 85: Hence also `sizehint!` might not be defined for `transitions`.

src/sample.jl
```diff
@@ -96,7 +96,9 @@ function mcmcsample(
             callback(rng, model, sampler, transition, i)
 
             # Save the transition.
-            transitions = save!!(transitions, transition, i, model, sampler, N; kwargs...)
+            oldT = typeof(transitions)
+            transitions = save!!(transitions, transition, i, model, sampler)
```
devmotion, line 100: We currently propagate all keyword arguments and the number of iterations all the time.

src/sample.jl
```diff
@@ -96,7 +96,9 @@ function mcmcsample(
...
+            oldT !== typeof(transitions) && Base.sizehint!(transitions, N)
```
devmotion, line 101: Again `sizehint!` might not be defined for `transitions`.

devmotion:
Do you have a MWE? IMO we should keep the interface and implementation in AbstractMCMC as general as possible, and as pointed out above the changes are quite restrictive. Maybe the underlying issue of the undefined reference could be fixed differently? Downstream packages can also always add their own implementation of `transitions` (hopefully only for samplers or models that they own...).

mohamed82008:
Here is a MWE:
```julia
julia> struct M <: AbstractMCMC.AbstractModel end

julia> struct S <: AbstractMCMC.AbstractSampler end

julia> a = Vector{Base.RefValue{Int}}(undef, 5); AbstractMCMC.save!!(a, Ref(1.0), 1, M(), S(), 5)
ERROR: UndefRefError: access to undefined reference
Stacktrace:
 [1] getindex at .\array.jl:787 [inlined]
 [2] copyto!(::Array{Base.RefValue,1}, ::Int64, ::Array{Base.RefValue{Int64},1}, ::Int64, ::Int64) at .\abstractarray.jl:841
 [3] append! at .\array.jl:955 [inlined]
 [4] copy! at .\abstractarray.jl:708 [inlined]
 [5] _setindex(::Array{Base.RefValue{Int64},1}, ::Base.RefValue{Float64}, ::Int64) at C:\Users\user\.julia\dev\BangBang\src\NoBang\base.jl:135
 [6] may at C:\Users\user\.julia\dev\BangBang\src\core.jl:10 [inlined]
 [7] setindex!! at C:\Users\user\.julia\dev\BangBang\src\base.jl:365 [inlined]
 [8] #save!!#11 at C:\Users\user\.julia\dev\AbstractMCMC\src\interface.jl:139 [inlined]
 [9] save!!(::Array{Base.RefValue{Int64},1}, ::Base.RefValue{Float64}, ::Int64, ::M, ::S, ::Int64) at C:\Users\user\.julia\dev\AbstractMCMC\src\interface.jl:139
 [10] top-level scope at REPL[23]:1
 [11] eval(::Module, ::Any) at .\boot.jl:331
 [12] eval_user_input(::Any, ::REPL.REPLBackend) at C:\Users\user\AppData\Local\Programs\Julia\Julia-1.4.0\share\julia\stdlib\v1.4\REPL\src\REPL.jl:86
 [13] run_backend(::REPL.REPLBackend) at C:\Users\user\.julia\packages\Revise\MgvIv\src\Revise.jl:1023
 [14] top-level scope at REPL[1]:0
```

devmotion:
Do you need to work with `Ref`? Or could you initialize the first transition in the correct way?

mohamed82008:
>Do you need to work with Ref? Or could you initialize the first transition in the correct way?

I don't work with `Ref` but with `Transition`. `Ref` is just a stand-in. Even if you initialize the first element properly, you get the same error when it tries to widen the type of the array because `copy!(b, a)` is called where `a` is the partially undefined vector which throws the above error. So we can either drop the `undef` definition or we use `push!!` instead of `setindex!!`.

devmotion:
> So we can either drop the undef definition or we use push!! instead of setindex!!.

Yeah, I guess we could just switch to `push!!` in general and get rid of the preallocation. That would also be closer to a transducer-based approach at some point (see https://github.com/TuringLang/AbstractMCMC.jl/pull/39).

Can you revert the changes and just use `push!!` in
https://github.com/TuringLang/AbstractMCMC.jl/blob/c7efc8a8beef89df8b4c045330dff69c0786627e/src/interface.jl#L139
(and adjust the docstring), change https://github.com/TuringLang/AbstractMCMC.jl/blob/c7efc8a8beef89df8b4c045330dff69c0786627e/src/interface.jl#L106 to
```julia
    t = Vector{typeof(transition)}(undef, 0)
    sizehint!(t, N)
    return t
```
and (while we're at it) https://github.com/TuringLang/AbstractMCMC.jl/blob/c7efc8a8beef89df8b4c045330dff69c0786627e/src/interface.jl#L115
to
```julia
    return Vector{typeof(transition)}(undef, 0)
```
? Then we keep the general interface but can optimize it still a bit for the default case of vectors.

mohamed82008:
Sounds good to me.

src/interface.jl
```diff
@@ -133,10 +135,13 @@ function save!!(
     iteration::Integer,
     ::AbstractModel,
     ::AbstractSampler,
-    ::Integer;
+    N::Integer;
     kwargs...
 )
-    return BangBang.setindex!!(transitions, transition, iteration)
+    @assert length(transitions) == iteration - 1
```
devmotion, line 141: This is not true in general (e.g., think about only saving every `n`th transition).
```suggestion
```

src/interface.jl
```diff
@@ -133,10 +135,13 @@ function save!!(
     iteration::Integer,
     ::AbstractModel,
     ::AbstractSampler,
-    ::Integer;
+    N::Integer;
```
devmotion, line 138: We don't need the number of iterations here in the default definition (users with custom implementations might want to use it):
```suggestion
    ::Integer;
```

src/interface.jl
```diff
@@ -133,10 +135,13 @@ function save!!(
...
     kwargs...
 )
-    return BangBang.setindex!!(transitions, transition, iteration)
+    @assert length(transitions) == iteration - 1
+    new_ts = BangBang.push!!(transitions, transition)
+    typeof(new_ts) !== typeof(transitions) && Base.sizehint!(new_ts, N)
+    return new_ts
```
devmotion, line 143: IMO the default implementation should only `push!!`:
```suggestion
    return BangBang.push!!(transitions, transition)
```

src/interface.jl
mohamed82008: This is needed for `sizehint!`.

src/interface.jl
```diff
@@ -103,7 +103,9 @@ function transitions(
     N::Integer;
     kwargs...
 )
-    return Vector{typeof(transition)}(undef, N)
+    ts = Vector{typeof(transition)}(undef, 0)
+    Base.sizehint!(ts, N)
```
devmotion, line 107: It's exported, isn't it?
```suggestion
    sizehint!(ts, N)
```

src/interface.jl
devmotion: But we don't want to use `sizehint!` here. We just push.

src/interface.jl
mohamed82008: `sizehint!` is needed to get similar performance to `setindex!!`.

src/interface.jl
```diff
@@ -147,11 +151,11 @@ function save!!(
     ::AbstractSampler;
     kwargs...
 )
+    @assert length(transitions) == iteration - 1
```
devmotion, line 154: Same here, this is not true in general:
```suggestion
```

src/interface.jl
devmotion: No, that happens once during initialization of `transitions`. There is no need to call `sizehint!` in every iteration.

mohamed82008: Only if the type changes because the output and input vectors would be different.

mohamed82008: The type will likely only change a few times so calling `sizehint!` here will only happen a few times.

devmotion: Ah I see, although I'm still not happy that about using `sizehint!` in this general definition. Think about the thinning example above, even if it is handled correctly during initialization we would suddenly preallocate much more memory than needed just because the type changed.

devmotion: So I think maybe rather downstream packages should have to pay the penalty for changing types or provide their custom implementation.

mohamed82008: Well the allocation was happening previously anyways. If we need to store less samples than `N` then maybe that should change in the `mcmcsample` function not `save!!`.

devmotion: The main idea behind the `save!!` functionality is to allow for a simple way to implement thinning without saving all transitions first or saving results to disk instead of storing them in memory.

But aside from that, I think it's probably not the best example since in that case you would implement `save!!` anyways for your own types. I'm just a bit worried about making any assumptions about how `transitions` is implemented in the default implementation of `save!!`. Users are supposed to be able to just change `transitions` if they want some other container type that maybe does not support `sizehint!`, without having to implement `save!!`. In the default implementation of `transitions` `sizehint!` makes sense since it returns a vector of transitions anyways, but IMO the natural default implementation of ``save!!`` would just use `push!!` - we just don't know if the user preallocated memory or wants to preallocate memory, and if yes, how much. So the simplest thing would be not to try to be clever at all and just use the same default as for the iterator-based sampling.

mohamed82008: So just because the default implementation uses a vector and `sizehint!` doesn't mean any other implementation should. `push!` may not even be used in the custom implementation. `N` and pre-allocation can be ignored in any implementation where it makes sense to ignore it. But for this specific implementation, it seems reasonable to use `sizehint!`. For thinning, won't you have to implement a different `mcmcsample` or `save!!` anyways? So then we can remove the `sizehint!` in that method, no?

devmotion: > For thinning, won't you have to implement a different mcmcsample or save!! anyways

You would not touch `mcmcsample` but implement a different `save!!` ideally (because otherwise all the default logic is gone). As I said it's not really an appropriate example in this context since you would implement `save!!` anyways.

A better example would be static arrays. With `push!!` you could actually save the transitions in static arrays but `sizehint!` is not defined for static arrays. The same should be true for any array type of fixed size.

If we don't care about breaking these cases for everyone we could do something like
- Define `save!!` only for `AbstractVector` container of transitions in AbstractMCMC
- Check if `newts === transitions` (that will check the types first but also make sure that the arrays are identical, since otherwise the size hint was lost as well probably) and apply `sizehint!` otherwise if the total number of iterations is provided.

However, I assume that depending on how often the type changes and how long the chains are, this could actually be less efficient than just pushing to the new vectors. And if the array type doesn't change the implementation without `sizehint!` would be very efficient since we call `sizehint!` when initializing the container in `transitions`. So I'm not sure there's much benefit in adding `sizehint!` here since it breaks some things that are currently possible and might still not be completely efficient (that would require some benchmarking though).

devmotion: BTW for thinning it would be much nicer to work with transducers or iterators I guess.

mohamed82008: I don't know why we may need to save the chains a static array. We can use `sizehint!` only for `AbstractVector`. That sounds fair enough.

>Check if newts === transitions (that will check the types first but also make sure that the arrays are identical, since otherwise the size hint was lost as well probably) and apply sizehint! otherwise if the total number of iterations is provided.

If the actual variable is changing but the type is the same, then that would be a bug in BangBang. If the type is not changing, there is no reason why we can't just `push!` normally. Anyways, this will be inside an if statement checking if `AbstractVector` or not. So it should be safe for methods of this function that don't use an `AbstractVector` for some reason. In fact, I would argue that using `BangBang` we are assuming that `transitions` is an instance of `AbstractVector` anyways so we might as well constraint the argument type. Overloading `save!!` for a custom `transitions` type makes more sense to me than overloading `BangBang.push!!`.

devmotion: > If the actual variable is changing but the type is the same, then that would be a bug in BangBang.

Not necessarily, you could have immutable containers whose size is not included in the type information.

I would be fine (but not completely satisfied :stuck_out_tongue:) with the suggestions in https://github.com/TuringLang/AbstractMCMC.jl/pull/40/files#r427928351, I guess it should be fine in most cases. If we experience any issues with `sizehint!` (bugs or performance), we can remove it.

PR #42, opened 2020-06-15
Switch design to immutable samplers

devmotion:
This PR is a proposal to fix https://github.com/TuringLang/AbstractMCMC.jl/issues/31.

The current interface with `sample_init!`, `step!`, `transitions`, `save!!`, `sample_end!`, and `bundle_samples` is replaced with
- `step(rng, model, sampler[, state; kwargs...])` that returns a 2-tuple of the next sample and state (similar to `Base.iterate`) and takes the state of the sampler as an argument in all but the initial step
- `samples(sample, model, sampler[, N; kwargs...])` creates a container for the samples with initial sample `sample` and (potentially) predefined number of samples `N`
- `save!!(samples, sample, iteration, model, sampler[, N; kwargs...])` saves the `sample` in the given `iteration` in the container of `samples`
- `bundle_samples(samples, model, sampler, state, chain_type[; kwargs...])` creates the returned chain of type `chain_type` from the `samples` (the final `state` of the sampler can be saved as well if desired)

Additionally, the PR contains some improvements (which ended up in here but could be moved to separate PRs if desired) such as
- avoiding hangs if the parallel sampling errors by adding a `try`-`finally` block (as suggested a while ago on Slack)
- removing some unneeded copies in the parallel sampling with multiple processes.


src/interface.jl
```diff
@@ -6,154 +6,100 @@ Concatenate multiple chains.
...
 """
-    step!(rng, model, sampler[, N = 1, transition = nothing; kwargs...])
-
-Return the transition for the next step of the MCMC `sampler` for the provided `model`,
-using the provided random number generator `rng`.
+    step(rng, model, sampler[, state; kwargs...])
 
-Transitions describe the results of a single step of the `sampler`. As an example, a
-transition might include a vector of parameters sampled from a prior distribution.
+Return a 2-tuple of the next sample and the next state of the MCMC `sampler` for `model`.
```
cpfiffer, line 32: How do you feel about encouraging or noting that people can use a NamedTuple instead? It's maybe a nice-to-have for dev types and it shouldn't screw anything up since you can index NamedTuples identically to tuples.

src/interface.jl
```diff
@@ -6,154 +6,100 @@ Concatenate multiple chains.
...
-`model`, whose first transition is `transition`.
+Generate a container for the samples of the MCMC `sampler` for the `model`, whose first
+sample is `sample`.
 
-The method can be called with and without a predefined size `N`.
+The method can be called with and without a predefined number `N` of samples.
 """
-function transitions(
-    transition,
+function samples(
```
cpfiffer, line 50: I'm a little worried about the `transitions` => `samples` name change, since I want to try to minimize the damage that downstream packages incur. They will already be doing (possibly) significant work to move to an immutable framework, and it'll be easier for them if function names don't change.

Additionally, while I definitely agree that `samples` is more informative, I think it's worth considering what a "weird" name like `transition` can do for people's packages. The nice thing about it is that they can define all kinds of types with names like `TuringTransition` or whatever that don't include the word `sample`, which appears in 90% of MCMC contexts and can mean different things each time. 

`transition` is nice because it doesn't collide with sampling in a semantic sense. Thoughts?

src/sample.jl
```diff
@@ -63,50 +63,44 @@ function mcmcsample(
...
 )
     # Check the number of requested samples.
     N > 0 || error("the number of samples must be ≥ 1")
 
-    # Perform any necessary setup.
-    sample_init!(rng, model, sampler, N; kwargs...)
-
     @ifwithprogresslogger progress name=progressname begin
-        # Obtain the initial transition.
-        transition = step!(rng, model, sampler, N; iteration=1, kwargs...)
```
cpfiffer, line 78: So we aren't providing information about what iteration we're in to the `step` call? I think it's relatively costless to pass `iteration` through.

src/interface.jl
devmotion: In general, I think there should be only one recommended way of doing things, so either one should return a tuple or a named tuple. IMO there's no clear advantage of returning a named tuple, also since it is just an internal method. Returning tuples is consistent with `Base.iterate` and if devs implement it incorrectly it will be apparent immediately from the resulting samples.

src/interface.jl
devmotion: I thought the name `transition` (as I understand it at least) is not correct here, since the function does not return a container for transitions to the next sampling step or states of the sampler for the next step, it's just a container of the samples. To me the term `samples` felt much more suitable and intuitive.

Most developers will not (and probably should not) actually implement `samples` (or the previous `transitions`), so I'm not sure how much breakage the rename would cause. That being said, we could easily add a deprecation for `transitions` and forward it to `samples` until the next major release.

If developers want to implement `samples`, it should not cause any namespace issues since it is not exported. Hence developers could just implement (and possibly call) `AbstractMCMC.samples` explicitly instead of importing it, which would not lead to any conflicts with other variables or functions named `samples`. Alternatively one could rename the function to something less generic such as `samples_container` (just the first thing that came to my mind).

src/sample.jl
devmotion: Yes, I agree, it's quite cheap to pass `iterations = i` around. However, actually I think it's a suboptimal approach. It is used by some samplers such as the particle filters that pop the samples from a precomputed list of samples, but with the clear distinction between state and samples it could now just be part of `state`. So first, again I think there should be only one way of doing things here, and removing `iteration = i` will avoid the question whether `iteration` should be part of the state or taken from the keyword argument, without sacrificing any functionality. Moreover, if `iteration` is part of the state everyone can just call `step!` in their own implementation of `sample` without having to worry to pass around `iteration = i` (since otherwise particle filters wouldn't be supported).

Along the same lines one could probably even remove `kwargs...` and make these options part of the sampler or model (whatever they are) since they have to be supported by the `step!` implementation anyways, but I didn't want to make even more drastic changes here for now.

src/interface.jl
cpfiffer: Good point, I'm happy with this.

src/interface.jl
cpfiffer: Let's just do a deprecation for now.

src/sample.jl
cpfiffer: I think it's really cheap to leave this in -- some developers can write their own iteration trackers and include it in the state, and that's fine. 

But I think it's worth including the ability to get a completely objective and fault-free iteration counter from the outer loop that a developer would otherwise not generally have access to. This is one of those basic record-keeping functions that we don't have to pass on to developers (by which I mean writing code to handle iterations as part of the state).

devmotion: I had the impression that the `iteration` keyword argument does not provide any additional information to the developer implementing `step!` for a certain model-sampler combination. Do you have any example in mind where that's not the case?

The `iteration` keyword argument felt wrong to me because `step!` implementations that rely on it will break as soon as it is not passed - which could easily happen for custom implementations of `sample` or in the iteration and transducer interface. If one tracks the iteration as part of the state (which is updated in every iteration step and hence should be equivalent?), then these problems do not exist.

cpfiffer: Okay, I'm fine with removing this.

devmotion:
I added a deprecation (and tests for it).

devmotion:
Please don't merge this PR, even though it's approved.

I'd like to have working branches of DynamicPPL and Turing available locally before merging this PR (should be ready within the next days).

cpfiffer:
Okay. Do you want to merge this into `dev` for now and then we can pull it into master when it's ready elsewhere?

devmotion:
I don't think there's an advantage in creating a `dev` branch and moving it there? I'd just keep the PR open for now (also as a reminder that there's still some work to do).

devmotion:
https://github.com/TuringLang/EllipticalSliceSampling.jl/compare/abstractmcmc2 shows the changes in EllipticalSliceSampling for this PR (also removed the `ESS_mcmc` method, by fully committing to `sample` we'll get the parallel sampling for free). Became much cleaner, now the hack with `transitions_init` and `transitions_save!` is not needed anymore - the interface already provides a clear separation between samples and states.

devmotion:
I'm wondering if we should actually just merge and release AbstractMCMC 2.0. There's no need to update all packages before it is released, with a new release I could just make a PR to EllipticalSliceSampling and run CI tests instead of relying on my local setup, and packages that are not compatible will be informed by CompatHelper about the new release but users will still only be able to use the version that was declared to be compatible.

cpfiffer:
Yeah, I think it's fine. I believe most everything downstream has the correct Semver bounds.

Issue #44, opened 2020-07-02
Default reduction for parallel sampling


devmotion:
I just ran into some annoying problem while adding some tests for parallel sampling in https://github.com/TuringLang/EllipticalSliceSampling.jl/pull/8.

In the simple test examples, the sampler just returns a vector of samples (the default for regular sampling) which are either scalars or vectors with one element. Unfortunately, parallel sampling does not work due to https://github.com/TuringLang/AbstractMCMC.jl/blob/f03a17d63fa794413cbf620ecc7ba46fba9b480b/src/sample.jl#L252-L253 (and similar lines) and `chaincat` being only defined for `AbstractChains`.

Now what should we do if the sampling does not return an object of type `AbstractChains`? Or what should we do in general?

Maybe the user could pass a `reduction` function which would default to `AbstractMCMC.DEFAULT_REDUCTION`, defined as
```julia
DEFAULT_REDUCTION(x::Vector{Any}) = map(identity, x)
DEFAULT_REDUCTION(x) = x
```
just to make sure that we return a concretely typed result if possible.

But of course that would mean that by default you would always get back a vector of chains (of whatever type), which would be quite breaking. But maybe that's not too bad and actually more intuitive?

cpfiffer:
I would make `chainscat(x::Vector{Any}...)` default to a vector-of-vectors. `chainscat(x::AbstractChains...)` should be required to implement a specific functional form. I don't think we need an additional reduction -- that's the purpose of `chainscat`.

devmotion:
But that would mean that we could never avoid splatting, even if we would like to use `reduce` since we will always have to split the vector of chains when we all `chainscat(...)`. Just defining `chainscat` on vectors doesn't seem to help either since then we can't distinguish between vectors of arrays and vectors of `AbstractChains` since in multithreaded sampling the vector is always of type `Vector{Any}`.

That's why I thought at some point, maybe we should just return the vectors and let users/downstream packages deal with the reduction (or possibly provide some reduction function directly).

francescoalemanno:
i stumbled upon this problem while implementing https://github.com/JuliaApproxInference/KissABC.jl/pull/16

devmotion:
Seems this was addressed in #45.

PR #45, opened 2020-07-02
Add `chainsstack`

devmotion:
This PR tries to address https://github.com/TuringLang/AbstractMCMC.jl/issues/44 by adding a method for (hopefully) concretizing arrays of type `Vector{Any}` and a `chainsstack(x::AbstractVector)` that falls back to `reduce(chainscat, x)` or `x`, depending on whether `eltype(x) <: AbstractChains` or not.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -1,9 +1,23 @@
 """
-    chainscat(c::AbstractChains)
+    chainscat(c::AbstractChains...)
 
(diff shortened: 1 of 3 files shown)


devmotion:
This helps to resolves issues with parallel sampling in EllipticalSliceSampling where each chain is just a vector of samples.

PR #46, opened 2020-07-03
Provide generic implementation of `bundle_samples` for `Vector{T}`

devmotion:
While updating AdvancedMH I noticed that we define a bunch of `bundle_samples` implementations there. I was wondering if we could generalize that in a way such that we don't have to repeat it in every package. In particular, it seems a bit strange to reimplement it in every package for `Vector{NamedTuple}` even though the only difference is how the samples are converted to a `NamedTuple`. But one could, e.g., also have a generic version for MCMCChains.Chains in MCMCChains with the basic structure + an interface for other packages to convert the samples, retrieve the names, the log-evidence + save the state, sampler, and model in a standardized way (and not different in every package).

This PR contains a very generic implementation for any desired vector-valued output.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -26,6 +26,19 @@ function bundle_samples(
     return samples
 end
 
+function bundle_samples(
+    samples::Vector,
+    ::AbstractModel,
+    ::AbstractSampler,
+    ::Any,
+    ::Type{Vector{T}};
+    kwargs...
+) where T
+    return map(samples) do sample
+        convert(T, sample)
+    end
+end
+
 """
     step(rng, model, sampler[, state; kwargs...])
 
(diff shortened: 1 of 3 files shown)


devmotion:
However, for AdvancedMH this would mean that the keyword argument `param_names` wouldn't be available anymore for `Vector{NamedTuple}` output. I'm wondering if that would be an issue? I guess users could still change the names afterwards (even for Chains) but requires an additional step?

devmotion:
Ah wait, I guess one could just define `convert(::Type{NamedTuple{names}}, ::Transition)` in NamedTuples and then use `Vector{NamedTuple{names}}` as output type :+1:

yebai:
@devmotion I fixed a merge conflict. Not sure it is what you intended to do. Pls, take a look anyway.

devmotion:
Thanks, looks good to me!

Issue #49, opened 2020-11-25
Option to thin


luiarthur:
Is there currently an option to thin samples by a certain factor? (I've looked in `sample.jl` and there doesn't seem to be such an option.)

I'm thinking of something like:
```julia
number_of_samples = 100
chain = sample(some_model, some_sampler, number_of_iterations; thin=3, discard_initial=50)
```
would run a Markov chain for 50 iterations (`discard_initial`), then another 300 iterations (`number_of_samples * thin`) but only every third (`thin`) sample is kept. Thus, 100 samples are returned.

This would be useful for certain samplers, like `MH`, where samples tend to be more correlated (than say those from ` NUTS`) and the user wants to keep only every `thin`-th sample so as to obtain less correlated samples while reducing memory consumption.

I'd be happy to implement this if there isn't currently an option to do so.

devmotion:
I guess it could be useful (I think we had this discussion at some point at least) but it hasn't been implemented yet mainly for two reasons: Turing does not use AbstractMCMC 2 but the old interface (this will change soon though) and we noticed that one has to be careful with changing things in AbstractMCMC since it affects many also Turing-independent downstream packages and hence it is not a good place to play around with ideas.

That being said, I think it could useful, not sure if it should be called `thin` or `thinning` or something else. Also at some point the future (not the current) code in Turing for resuming sampling should be moved to AbstractMCMC I assume.

By the way, if you are concerned about memory, you can use the iteration and transducer interfaces.

luiarthur:
Yes, I recall having that conversation but couldn't find the thread...

> one has to be careful with changing things in AbstractMCMC since it affects many also Turing-independent downstream packages and hence it is not a good place to play around with ideas.

Makes sense. Is this then perhaps something that's more suited to be implemented in Turing? Though, my naive view is that this is mainly an accounting problem that involves something like the following change in `mcmcsample` (and perhaps other relevant places):

```julia
    Ntotal = N * thin + discard_initial

    # stuff ...

        # Step through the sampler.
        for i in 2:(N*thin)
            # Obtain the next sample and state.
            sample, state = step(rng, model, sampler, state; kwargs...)

            # Run callback.
            callback === nothing || callback(rng, model, sampler, sample, i)

            # Save the sample.
            (i % thin == 0) && (samples = save!!(samples, sample, i, model, sampler, N; kwargs...))

            # Update the progress bar.
[...]

PR #50, opened 2020-11-29
Implemented `thinning` option in `mcmcsample`

luiarthur:
I implemented a `thinning` option in `mcmcsample` so that a user can thin samples by a certain factor, in reference to #49.

I'm not sure what else to check besides that the number of samples returned is correct. I'd be happy add more tests and documentation as directed.

Also, I'm totally fine if this is rejected since @devmotion mentioned that there could be unintended effects on Turing-independent packages that rely on AbstractMCMC. It's just something that I and (I think other) people that use `GibbsConditional` and `MH` might find useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,12 +72,7 @@ are:
   `callback(rng, model, sampler, sample, iteration)` is called after every sampling step,
   where `sample` is the most recent sample of the Markov chain and `iteration` is the current iteration
 - `discard_initial` (default: `0`): number of initial samples that are discarded
-- `thinning` (default: `1`): factor by which to thin samples. E.g., if `number_of_samples=100`,
-  then `sample(some_model, some_sampler, number_of_samples; thinning=3, discard_initial=50)`
-  would run a Markov chain for 50 iterations (`discard_initial`), then another
-  300 iterations (`number_of_samples * thinning`), but only every third
-  (`thinning`) sample is kept. Thus, 350 transitions take place, but only 100
-  samples are returned. 
+- `thinning` (default: `1`): factor by which to thin samples.
 
 Additionally, AbstractMCMC defines the abstract type `AbstractChains` for Markov chains and the
 method `AbstractMCMC.chainscat(::AbstractChains...)` for concatenating multiple chains.
--- a/README.md
+++ b/README.md
@@ -72,6 +72,12 @@ are:
   `callback(rng, model, sampler, sample, iteration)` is called after every sampling step,
   where `sample` is the most recent sample of the Markov chain and `iteration` is the current iteration
 - `discard_initial` (default: `0`): number of initial samples that are discarded
+- `thinning` (default: `1`): factor by which to thin samples. E.g., if `number_of_samples=100`,
+  then `sample(some_model, some_sampler, number_of_samples; thinning=3, discard_initial=50)`
+  would run a Markov chain for 50 iterations (`discard_initial`), then another
+  300 iterations (`number_of_samples * thinning`), but only every third
+  (`thinning`) sample is kept. Thus, 350 transitions take place, but only 100
+  samples are returned. 
 
 Additionally, AbstractMCMC defines the abstract type `AbstractChains` for Markov chains and the
 method `AbstractMCMC.chainscat(::AbstractChains...)` for concatenating multiple chains.
--- a/README.md
+++ b/README.md
@@ -68,9 +68,10 @@ Common keyword arguments for regular and parallel sampling (not supported by the
 are:
 - `progress` (default: `true`):  toggles progress logging
 - `chain_type` (default: `Any`): determines the type of the returned chain
-- `callback` (default: `nothing`): if `callback !== nohting`, then
+- `callback` (default: `nothing`): if `callback !== nothing`, then
   `callback(rng, model, sampler, sample, iteration)` is called after every sampling step,
   where `sample` is the most recent sample of the Markov chain and `iteration` is the current iteration
+- `discard_initial` (default: `0`): number of initial samples that are discarded
 
 Additionally, AbstractMCMC defines the abstract type `AbstractChains` for Markov chains and the
 method `AbstractMCMC.chainscat(::AbstractChains...)` for concatenating multiple chains.
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -114,7 +114,7 @@ function mcmcsample(
             # Run callback.
             callback === nothing || callback(rng, model, sampler, sample, i)
 
-            # Save every `thinning`-th sample.
+            # Save the sample.
             samples = save!!(samples, sample, i, model, sampler, N; kwargs...)
 
             # Update the progress bar.
@@ -188,7 +188,7 @@ function mcmcsample(
             # Run callback.
             callback === nothing || callback(rng, model, sampler, sample, i)
 
-            # Save every `thinning`-th sample.
+            # Save the sample.
             samples = save!!(samples, sample, i, model, sampler; kwargs...)
 
             # Increment iteration counter.
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -71,7 +71,7 @@ function mcmcsample(
 )
     # Check the number of requested samples.
     N > 0 || error("the number of samples must be ≥ 1")
-    Ntotal = thinning * N + discard_initial
+    Ntotal = thinning * (N - 1) + discard_initial + 1
 
     @ifwithprogresslogger progress name=progressname begin
         # Obtain the initial sample and state.
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -105,7 +105,10 @@ function mcmcsample(
                 sample, state = step(rng, model, sampler, state; kwargs...)
                 
                 # Update progress bar.
-                progress && ProgressLogging.@logprogress (itotal += 1) / Ntotal
+                if progress
+                    itotal += 1
+                    ProgressLogging.@logprogress itotal / Ntotal
+                end
             end
 
             # Obtain the next sample and state.
@@ -118,7 +121,10 @@ function mcmcsample(
             samples = save!!(samples, sample, i, model, sampler, N; kwargs...)
 
             # Update the progress bar.
-            progress && ProgressLogging.@logprogress (itotal += 1) / Ntotal
+            if progress
+                itotal += 1
+                ProgressLogging.@logprogress itotal / Ntotal
+            end
         end
     end
 
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -97,7 +97,16 @@ function mcmcsample(
         progress && ProgressLogging.@logprogress (1 + discard_initial) / Ntotal
 
         # Step through the sampler.
-        for i in 2:(thinning * N)
+        itotal = 1 + discard_initial
+        for i in 2:N
+            # Discard thinned samples.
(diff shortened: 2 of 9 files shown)


src/sample.jl
```diff
@@ -96,15 +97,17 @@ function mcmcsample(
...
 
-            # Save the sample.
-            samples = save!!(samples, sample, i, model, sampler, N; kwargs...)
+            # Save every `thinning`-th sample.
+            if i % thinning == 1 || thinning == 1
+              samples = save!!(samples, sample, i, model, sampler, N; kwargs...)
+            end
 
             # Update the progress bar.
             progress && ProgressLogging.@logprogress (i + discard_initial) / Ntotal
```
devmotion, line 113:
```suggestion
            # Save every `thinning`-th sample.
            samples = save!!(samples, sample, i, model, sampler, N; kwargs...)

            # Update the progress bar.
            progress && ProgressLogging.@logprogress (itotal += 1) / Ntotal
```

src/sample.jl
```diff
@@ -170,8 +174,10 @@ function mcmcsample(
             # Run callback.
             callback === nothing || callback(rng, model, sampler, sample, i)
 
-            # Save the sample.
-            samples = save!!(samples, sample, i, model, sampler; kwargs...)
+            # Save every `thinning`-th sample.
+            if i % thinning == 1 || thinning == 1
+              samples = save!!(samples, sample, i, model, sampler; kwargs...)
+            end
```
devmotion, line 180: The logic should be changed similar to the suggestion above.

test/sample.jl
```diff
@@ -226,6 +226,14 @@
         @test !ismissing(chain[1].a)
     end
 
+    @testset "Thin chain by a factor of `thinning`" begin
+        chain = sample(MyModel(), MySampler(), 100; sleepy = true, discard_initial = 50,
+                       thinning = 3)
```
devmotion, line 231:
```suggestion
        chain = sample(MyModel(), MySampler(), 100; sleepy = true, thinning = 3)
```

test/sample.jl
```diff
@@ -238,6 +246,12 @@
         bmean = mean(x.b for x in chain)
         @test !ismissing(chain[1].a)
         @test abs(bmean) <= 0.001 && length(chain) < 10_000
+
+        # Thin chain by a factor of `thinning`..
+        chain = sample(MyModel(), MySampler(); discard_initial = 50, thinning = 3)
```
devmotion, line 251:
```suggestion
        chain = sample(MyModel(), MySampler(); thinning = 3)
```

devmotion (changes requested):
Thanks for the PR! I added some comments and suggestions. Regarding the tests: one could/should maybe just sample twice with the same seed, e.g., once with `thinning = 1` and `3 * nsamples` and once with `thinning = 1` and `nsamples` samples and check that the correct samples are returned. However, we also have to switch to Github Actions first.

luiarthur:
Thanks for the comments!

> Regarding the tests: one could/should maybe just sample twice with the same seed, e.g., once with `thinning = 1` and `3 * nsamples` and once with `thinning = 1` and `nsamples` samples and check that the correct samples are returned. 

I think I understand what you're saying here. Do you mean "once with `thinning = 1` and `3 * nsamples` and once with `thinning = 3` and `nsamples` samples", to check that the last sample is the same for both chains?

README.md
```diff
@@ -72,6 +72,12 @@ are:
   `callback(rng, model, sampler, sample, iteration)` is called after every sampling step,
   where `sample` is the most recent sample of the Markov chain and `iteration` is the current iteration
 - `discard_initial` (default: `0`): number of initial samples that are discarded
+- `thinning` (default: `1`): factor by which to thin samples. E.g., if `number_of_samples=100`,
+  then `sample(some_model, some_sampler, number_of_samples; thinning=3, discard_initial=50)`
+  would run a Markov chain for 50 iterations (`discard_initial`), then another
+  300 iterations (`number_of_samples * thinning`), but only every third
+  (`thinning`) sample is kept. Thus, 350 transitions take place, but only 100
+  samples are returned.
```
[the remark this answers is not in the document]
luiarthur: Thanks for the comments!

> Regarding the tests: one could/should maybe just sample twice with the same seed, e.g., once with `thinning = 1` and `3 * nsamples` and once with `thinning = 1` and `nsamples` samples and check that the correct samples are returned. 

I think I understand what you're saying here. Do you mean "once with `thinning = 1` and `3 * nsamples` and once with `thinning = 3` and `nsamples` samples", to check that the third-to-the-last sample in the first chain is the same as the last sample in the second chain?

devmotion:
Yes, the idea is to check that the samples 1, k + 1, 2 * k + 1, ..., (nsamples - 1) * k + 1 in the chain without thinning are the same as samples 1, 2, ..., nsamples in the chain with `thinning = k`.

src/sample.jl
```diff
@@ -96,15 +97,17 @@ function mcmcsample(
         progress && ProgressLogging.@logprogress (1 + discard_initial) / Ntotal
 
         # Step through the sampler.
-        for i in 2:N
+        for i in 2:(thinning * N)
```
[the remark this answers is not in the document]
luiarthur: Thanks for the keen eye. Yes, the suggestion here and below makes sense. I have a question below.

src/sample.jl
luiarthur: Is updating the progress bar in line 111 necessary, since it would already be updated while thinning?

devmotion: It is updated every time a new sample is obtained. In the thinning loop `thinning - 1` new samples are obtained, and afterwards one additional sample which is also saved then.

luiarthur: My bad. Just realized the suggestion reads `for _ in 1:(thinning - 1)`.

cpfiffer:
I have nothing to add on top of David's suggestions -- great work! I actually really like this in AbstractMCMC. It's exactly the kind of logic that can be centralized.

luiarthur:
Thanks for the clarification @devmotion. I've made the changes and added a test to ensure that a thinned chain is every k-th sample of a longer (un-thinned) chain.

test/sample.jl
```diff
@@ -226,6 +226,27 @@
...
+
+        # Run a long chain with N * K samples.
+        Random.seed!(1234)
+        long_chain = sample(MyModel(), MySampler(), N * K; sleepy = true)
+        @test length(long_chain) == N * K
+        @test ismissing(long_chain[1].a)
+
+        # Check that thinned_chain is every k-th sample in long_chain.
+        # NOTE: `i` begins at 2 since first sample in `thinned_chain` is `missing`.
+        @test all([thinned_chain[i].a == long_chain[(i-1) * K + 1].a for i in 2:N])
```
devmotion, line 246: I'm a bit picky here but I think it would be nice to use a more descriptive name such as `thinning` instead of `K`. We also don't have to test the reference chain, we just compare with it:
```suggestion
        # Run a thinned chain with `N` samples thinned by factor of `thinning`.
        Random.seed!(1234)
        N = 100
        thinning = 3
        chain = sample(MyModel(), MySampler(), N; sleepy = true, thinning = thinning)
        @test length(chain) == N
        @test ismissing(chain[1].a)

        # Repeat sampling without thinning.
        Random.seed!(1234)
        ref_chain = sample(MyModel(), MySampler(), N * thinning; sleepy = true)
        @test all(chain[i].a === ref_chain[(i - 1) * thinning + 1].a for i in 1:N)
```

devmotion:
Looks good :+1: We just have to wait for https://github.com/TuringLang/AbstractMCMC.jl/pull/51 and then make sure that tests pass. Can you already bump the version to 2.2.0?

test/sample.jl
luiarthur: I haven't a strong opinion here. I'll just note that "thinning" is a rather common term for Bayesian modelers.

As for alternate names, STAN uses `thin`, tensorflow-probability uses `num_steps_between_results`. Do you have a preference or another term in mind.

luiarthur: Sorry, I misread again. I guess you were referring to `K` in the test only? Yes, let me change that.

luiarthur:
Thanks! Yes, just changed the version to 2.2.0 in `Project.toml`.

devmotion:
There seems to be a problem with the progress bars, e.g., the logs show progress bars that go up to 200%.

src/sample.jl
```diff
@@ -65,12 +65,13 @@ function mcmcsample(
...
     callback = nothing,
     discard_initial = 0,
+    thinning = 1,
     chain_type::Type=Any,
     kwargs...
 )
     # Check the number of requested samples.
     N > 0 || error("the number of samples must be ≥ 1")
-    Ntotal = N + discard_initial
+    Ntotal = thinning * N + discard_initial
```
devmotion, line 74: I think this is not correct and causes the incorrect progress bars:
```suggestion
    Ntotal = thinning * (N - 1) + discard_initial + 1
```

luiarthur: Ah. That makes sense because the last few iterations aren't saved. Thanks!

devmotion:
The progress bars are still incorrect.

devmotion:
Maybe there's something strange going on with the `@logprogress` macro here. Can you try to replace all occurrences of
```julia
progress && @logprogress (itotal += 1) / Ntotal
```
with
```julia
if progress
    itotal += 1
    @logprogress itotal / Ntotal
end
```
?

luiarthur:
Yes, I'll try that.

luiarthur:
I apologize for not catching this. When I ran the test in the julia repl, the progress bar advanced so quickly that I only see the final state, which for some reason always shows up as 100%...

Anyway, your suggestion to replace
```julia
progress && @logprogress (itotal += 1) / Ntotal
```
with
```julia
if progress
    itotal += 1
    @logprogress itotal / Ntotal
end
```
fixed the problem. 

But it seems that both:
```julia
Ntotal = thinning * (N - 1) + discard_initial + 1
```
and
```julia
Ntotal = thinning * N + discard_initial
```
work.

I haven't committed anything yet. But I thought about this a bit more, and it seems the latter should be correct. Any thoughts?

devmotion:
> I haven't committed anything yet. But I thought about this a bit more, and it seems the latter should be correct. Any thoughts?

The latter is not correct: we generate `discard_initial` samples that are discarded, then one sample that is saved, and afterwords we run the for-loop `N - 1` times and generate `thinning` samples in each run. So in total `discard_initial + 1 + (N - 1) * thinning` samples are generated.

However, the difference is not noticeable in our tests: the difference between both definitions of `Ntotal` is `thinning - 1` (less in the correct case). I.e., in all tests with a default value of `thinning = 1`, they yield exactly the same `Ntotal` value, and in the tests with thinning we use `thinning = 3`, i.e., the difference is 2. Since we generate 1000 samples, this difference is not noticeable.

luiarthur:
Yes. Thanks for the correction. Just pushed the correction.

devmotion (approved):
Seems fixed now, the incorrect progress bars are gone.

Thanks again for the PR!

luiarthur:
Awesome, thank! Looks like #51 was merged. Do I now merge this PR?

devmotion:
If tests pass, you can merge it. Otherwise, if you do not have sufficient permissions, @cpfiffer or I will merge the PR once tests have passed.

PR #51, opened 2020-11-29
Add Github CI

devmotion:


.github/workflows/CompatHelper.yml
```diff
@@ -14,4 +12,5 @@ jobs:
       - name: CompatHelper.main()
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMPATHELPER_PRIV: ${{ secrets.COMPATHELPER_PRIV }}
```
devmotion, line 15: This secret has to be added: https://github.com/JuliaRegistries/CompatHelper.jl#12-set-up-the-ssh-deploy-key-optional

cpfiffer: I think the trend here is towards using Github for all our repos, so we should probably just use an organization-wide secret.

@yebai can you make me a TuringLang admin for like a day or something so I can set up the keys? It should only take me like 5 minutes total.

devmotion: Yes, that's probably even better 👍

cpfiffer: I added a repo-specific token for the moment, so it should work.

devmotion: Can you approve if you think that the PR is ready to be merged? Then we could run the tests in the other PR.

cpfiffer (approved):

PR #53, opened 2020-12-17
Fixed missing semi-colon for kwargs in `steps`

torfjelde:
This really seems like a bug to me, but maybe I'm just brainfarting at 1am (it has happened before).


devmotion (approved):
I agree, this seems incorrect. I'm surprised that it is not covered by the tests. Maybe add a test with some keyword arguments?

In any case, can you update the version number as well for a bugfix release?

torfjelde:
Should be good now :+1:

Issue #55, opened 2021-02-25
Passing state + kwargs to callback


theogf:
Is there any reason for the `callback` function to not pass the `state` as well the additional `kwargs...` passed to sample?

The motivation for `state` is obvious (I suppose) and for `kwargs...` it would allow to pass storage/testing variables more easily.
For example :
`mycallback(rng, model, sampler, sample; X_test) = my_test(sample, X_test)`

Happy to make a PR if this would be an OK thing to add.

EDIT: Note that this would be breaking.

cpfiffer:
I think it's probably just a minor oversight that it's not doing this already. I'd be fine doing this.

PR #56, opened 2021-02-25
Include `state` and `kwargs...` to `callback`

theogf:
Following #55 , `state` and `kwargs...` are passed to the `callback` function.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "2.2.1"
+version = "3.0.0"
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ are:
 - `progress` (default: `true`):  toggles progress logging
 - `chain_type` (default: `Any`): determines the type of the returned chain
 - `callback` (default: `nothing`): if `callback !== nothing`, then
-  `callback(rng, model, sampler, sample, iteration)` is called after every sampling step,
+  `callback(rng, model, sampler, sample, state, iteration; kwargs...)` is called after every sampling step,
   where `sample` is the most recent sample of the Markov chain and `iteration` is the current iteration
 - `discard_initial` (default: `0`): number of initial samples that are discarded
 - `thinning` (default: `1`): factor by which to thin samples.
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -87,7 +87,7 @@ function mcmcsample(
         end
 
         # Run callback.
-        callback === nothing || callback(rng, model, sampler, sample, 1)
+        callback === nothing || callback(rng, model, sampler, sample, state, 1; kwargs...)
 
         # Save the sample.
         samples = AbstractMCMC.samples(sample, model, sampler, N; kwargs...)
@@ -115,7 +115,7 @@ function mcmcsample(
             sample, state = step(rng, model, sampler, state; kwargs...)
 
             # Run callback.
-            callback === nothing || callback(rng, model, sampler, sample, i)
+            callback === nothing || callback(rng, model, sampler, sample, state, i; kwargs...)
 
             # Save the sample.
             samples = save!!(samples, sample, i, model, sampler, N; kwargs...)
--- a/test/sample.jl
+++ b/test/sample.jl
@@ -276,4 +276,14 @@
         @test mean(x.b for x in chain) ≈ 0 atol=0.1
         @test var(x.b for x in chain) ≈ 1 atol=0.15
     end
+    
+    @testset "Testing callbacks" begin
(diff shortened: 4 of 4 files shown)


devmotion:
I assume that `state` could be useful potentially, even though I haven't encountered an example myself and it is a sampler-specific internal object. I assume that it was just not included since there has not been a use case and in contrast to the samples it is not user-facing and its structure can be completely arbitrary.

I am a bit more reluctant with the keyword arguments though. They were not included on purpose and I think there are good reasons for keeping it this way. If we include them here, then every callback has to handle keyword arguments (most by just discarding `kwargs...`, I assume). It seems most (all?) cases could be handled by including them in the callback on the call site instead, so I am not sure if the increased flexibility here outweighs the fact that every callback implementation has to account for keyword arguments. Do you have any concrete application in mind that can't be handled with the current setup?

theogf:
Ok so here is my example which might makes things more clear.
I work on https://github.com/theogf/BayesianQuadrature.jl, which basically does things in two steps:
- Draw samples (this uses the AbstractMCMC interface)
- Compute the quadrature

Now of course it would be nice to have a callback during the sampling to estimate the quadrature after every draw. Except that one needs the whole collection of samples to do that (this actually makes me realize that passing `samples` would be more interesting than passing `sample`).
I did not think it through completely and added `state` because in some of my samplers, the collection of samples is anyway saved in `state`.

About the keyword argument:
You can indeed have something like:
```
my_array = []
my_callback(args...) = push!(my_array, i)
```
but if you don't work inside a closure it means you need to work with global variables, having a simple alternative would be nice.

> outweighs the fact that every callback implementation has to account for keyword arguments

I don't buy this argument since you need to impose a structure for the callback function anyway...
One needs to (currently) write `callback` as `callback(rng, model, sampler, sample, i)` but with `kwargs...` the constraint is just a different one.

devmotion:
> Now of course it would be nice to have a callback during the sampling to estimate the quadrature after every draw. 

Maybe it would make sense to use the iteration (or transducer) interface in this case. Of course, some of the defaults such as progress bars etc. would be lost (I don't know how important they are for you) but you would have direct access to all samples and could easily collect them, perform additional computations (also with the collection of samples), apply "callbacks" only in every n-th step etc.

> but if you don't work inside a closure it means you need to work with global variables, having a simple alternative would be nice.

Hmm I'm not sure what exactly you mean - are you referring to the fact that one should avoid to clse over non-constant global variables? In this case one can apply e.g. the `let` block trick.

> I don't buy this argument since you need to impose a structure for the callback function anyway...

True, it's just a slightly more complex structure. But why make it more complex if it is not needed? In other places we actually reduced the number of positional and keyword arguments since they were not used and made the interface more complex. IMO it is good to keep the interface as simple as possible (but not simpler). Though maybe this does not apply to keyword arguments to the same extent as positional arguments since they can be dropped by developers quite easily in the implementation with the catch-all `kwargs...` whereas positional arguments have to be specified.

cpfiffer:
I think it kind of depends on how powerful we want the callbacks to be. The current format is rigorous and simple, but not as extensible as if keyword arguments were included.

I'm on the side of adding keyword arguments mostly because devs are going to have to make a change anyway to accommodate the `state` argument and can add `; kwargs...` to their callbacks quite easily. 

That said, I'm sympathetic to the argument that that is kind of an annoying thing to do if you are writing a quick-and-dirty callback function. Maybe we can catch this internally and allow both forms. Is there a robust way to check if the signature of the callback accepts keyword arguments, other than `try`/`catch`?

theogf:
I find the callback problem generally so annoying!
That would be nice to have a standard way to do this with the least restrictions for both the devs and the users. Is there something like this already?

I was thinking: how about passing everything via keywords arguments?
Like any callback function should be of the form `my_callback(;kwargs...)` which can then be refined `my_callback(;model, kwargs...)` etc and internally the callback function is always `callback(;model=model, state=state, i=i, etc..., kwargs...)`

torfjelde:
Just to rejuvenate this discussion: I also want `kwargs` :upside_down_face: 

> Though maybe this does not apply to keyword arguments to the same extent as positional arguments since they can be dropped by developers quite easily in the implementation with the catch-all kwargs... whereas positional arguments have to be specified.

And this is why I'm pro kwargs despite slight increase in complexity. Current moving AHMC over to AMCMC interface, and I think we want to continue using the process tracking that already exists in AHMC (due to additional information provided). AFAIK it's easiest to implement this is as a `callback`, BUT the issue is that I want to know stuff like `nadapts`, if the user specified `verbose`, etc. so that I can inform the user that "hey, adaptation just completed!" in the `callback`. 

Of course, all of this is easy to implement using the iterator or transducer interface, but IMO it's somewhat breaking with "user expectation": I very much *expected* `kwargs...` to be available (especially given that's being passed around to everything else in `sample` :sweat_smile: ).

cpfiffer:
Okay, I think I'm perfectly happy to add keyword arguments.

> EDIT: Ughh, I also need `state` disappointed

Yeah, I think this is a non-negotiable add -- `state` definitely needs to be passed on.

cpfiffer:
@theogf could you bump the version to 3.0, since this will be very breaking?

devmotion (approved):

cpfiffer:
Nevermind, I changed the version myself.

PR #59, opened 2021-03-14
Allow to enable and disable progress logging globally

devmotion:
This PR moves the `setprogress!` logic from Turing upstream. IMO it belongs here since it allows to tune the default keyword arguments. The change is non-breaking since the default is not changed and it can be specified in the `sample`/`mcmcsample` calls as before.

Some additional background:
The main reason for `mcmcsample` is that Turing wants to use different default keyword arguments for `sample` but I guess it might be better to allow to set or define the keyword arguments instead. E.g., by adding an additional `default_chain_type(model, sampler) = Any` function that could be implemented as `default_chain_type(::Model, ::Sampler{<:InferenceAlgorithm}) = Chain` in Turing.


cpfiffer:
I'm all for it, thanks!

PR #60, opened 2021-03-19
Limit update frequency of progress bars

devmotion:
This PR limits the update frequency of the progress bars to one update per 0.5% progress. This is the default value of [`@progress`](https://junolab.org/ProgressLogging.jl/dev/#ProgressLogging.@progress-Tuple). Progress logging is quite slow and therefore this should speed up sampling.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "2.3.0"
+version = "2.4.0"
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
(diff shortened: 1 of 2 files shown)


devmotion:
I'm a bit worried about this error: https://github.com/TuringLang/AbstractMCMC.jl/pull/60/checks?check_run_id=2146808318#step:6:468

cpfiffer:
Uh yeah that's a weird one. Seems to be related to (https://github.com/JuliaLang/julia/issues/35689)?

devmotion:
I still don't understand which change could cause it (provided it is the first time it appeared - since tests don't fail we might have missed it before). I've never seen this error before but it seems it showed up in CI test of MLJ as well: https://github.com/alan-turing-institute/MLJTuning.jl/pull/49#issuecomment-640089283

devmotion:
It seems the errors disappeared. Not sure if the commit was helpful though.

cpfiffer:
Maybe? Dunno. Either way it's good to go.

PR #61, opened 2021-03-19
Use TerminalLoggers also on Windows for Julia >= 1.5.3

devmotion:
It seems TerminalLoggers can be used for rendering progress bars on Windows with recent versions of Julia: https://c42f.github.io/TerminalLoggers.jl/dev/#Progress-bars-1

I don't have access to a computer with Windows, so it would be good if someone could confirm that it works fine before merging the PR.


cpfiffer:
I don't have windows either! Might just merge this through anyways since the change is fairly minor.

cpfiffer (approved):

PR #64, opened 2021-04-01
Add Metadata struct

cpfiffer:
This PR adds a `Metadata` struct that tracks the following information:

1. Start time
2. Stop time
3. Step time (time spent per step)
4. Step calls (how many times `step!` was called)
5. Allocations

It's created and passed as a keyword argument to `bundle_samples`, so it remains up to end-chain software to determine what to do (if anything) with the metadata.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "2.5.0"
+version = "2.5.1"
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
--- a/Project.toml
+++ b/Project.toml
@@ -8,6 +8,7 @@ version = "2.5.0"
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
 ConsoleProgressMonitor = "88cd18e8-d9cc-4ea6-8889-5259c0d15c8b"
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 LoggingExtras = "e6f89c97-d47a-5376-807f-9c37f3926c36"
--- a/Project.toml
+++ b/Project.toml
@@ -8,7 +8,6 @@ version = "2.6.0"
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
 ConsoleProgressMonitor = "88cd18e8-d9cc-4ea6-8889-5259c0d15c8b"
-Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 LoggingExtras = "e6f89c97-d47a-5376-807f-9c37f3926c36"
--- a/src/metadata.jl
+++ b/src/metadata.jl
@@ -24,7 +24,7 @@ end
 Metadata() = Metadata(Dates.now(), missing, 0,0,0)
 
 function update(f, md::Metadata)
-    value, stats... = @timed f(md)
+    (value, stats...) = @timed f(md)
 
     md.step_time += stats.time
     md.step_calls += 1
--- a/src/metadata.jl
+++ b/src/metadata.jl
@@ -1,6 +1,18 @@
 using Dates
 
-# Tracks sampling metadata
+"""
+    Metadata
+
+A struct that tracks sampling information. 
+
+The fields available are:
+
+- `start_time`: A `DateTime` indicating when sampling begun.
+- `stop_time`: A `DateTime` indicating when sampling finished.
+- `step_calls`: The number of times `step!` was called.
+- `step_time`: The total number of seconds spent inside `step!` calls.
+- `allocations`: The total number of bytes allocated by `step!`.
+"""
 mutable struct Metadata
     start_time::DateTime
     stop_time::Union{DateTime, Missing}
--- a/src/metadata.jl
+++ b/src/metadata.jl
@@ -24,11 +24,11 @@ end
 Metadata() = Metadata(Dates.now(), missing, 0,0,0)
 
 function update(f, md::Metadata)
-    (value, stats...) = @timed f(md)
+    value, etime, alloc, gct, _ = @timed f(md)
 
-    md.step_time += stats.time
+    md.step_time += etime
     md.step_calls += 1
-    md.allocations += stats.bytes
+    md.allocations += alloc
 
     return value
 end
--- a/src/metadata.jl
+++ b/src/metadata.jl
@@ -24,13 +24,13 @@ end
 Metadata() = Metadata(Dates.now(), missing, 0,0,0)
 
 function update(f, md::Metadata)
-    value, etime, alloc, gct, _ = @timed f(md)
+    (sample, state), etime, alloc, gct, _ = @timed f(md)
 
     md.step_time += etime
     md.step_calls += 1
     md.allocations += alloc
 
-    return value
+    return (sample, state)
 end
 
 function stop(md::Metadata)
--- a/src/metadata.jl
+++ b/src/metadata.jl
@@ -1,0 +1,26 @@
+using Dates
+
+# Tracks sampling metadata
+mutable struct Metadata
+    start_time::DateTime
+    stop_time::Union{DateTime, Missing}
+    step_time::Float64
+    step_calls::Int64
+    allocations::Int64
+end
+
+Metadata() = Metadata(Dates.now(), missing, 0,0,0)
+
+function update(f, md::Metadata)
+    value, stats... = @timed f(md)
+
+    md.step_time += stats.time
+    md.step_calls += 1
+    md.allocations += stats.bytes
+
+    return value
+end
+
+function stop(md::Metadata)
+    md.stop_time = Dates.now()
+end
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -87,6 +87,7 @@ function mcmcsample(
 
     # Start the timer
     start = time()
+    local state
 
     allocations = @allocated begin
         @ifwithprogresslogger progress name=progressname begin
@@ -207,6 +208,7 @@ function mcmcsample(
 
     # Start the timer
     start = time()
+    local state
 
     allocations = @allocated begin
         @ifwithprogresslogger progress name=progressname begin
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -79,12 +79,16 @@ function mcmcsample(
     discard_initial = 0,
     thinning = 1,
     chain_type::Type=Any,
-    sampling_metadata = Metadata(),
     kwargs...
 )
     # Check the number of requested samples.
     N > 0 || error("the number of samples must be ≥ 1")
     Ntotal = thinning * (N - 1) + discard_initial + 1
+
+    # Set up sampling stats -- timing, allocations, etc.
+    start = time()
+    step_time = 0
+    allocations = 0
 
     @ifwithprogresslogger progress name=progressname begin
         # Determine threshold values for progress logging
@@ -95,10 +99,10 @@ function mcmcsample(
         end
 
         # Obtain the initial sample and state.
-        sample, state = update(sampling_metadata) do x
-            step(rng, model, sampler; kwargs...)
-        end
-
+        (sample, state), _time, _allocations, _, _ = 
+            @timed step(rng, model, sampler; kwargs...)
+        step_time += _time
+        allocations += _allocations
 
         # Discard initial samples.
         for i in 1:(discard_initial - 1)
@@ -109,9 +113,10 @@ function mcmcsample(
             end
 
             # Obtain the next sample and state.
-            sample, state = update(sampling_metadata) do x
-                step(rng, model, sampler, state; kwargs...)
-            end
+            (sample, state), _time, _allocations, _, _ =  
+                @timed step(rng, model, sampler, state; kwargs...)
+            step_time += _time
+            allocations += _allocations
         end
 
         # Run callback.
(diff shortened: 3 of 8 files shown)


Project.toml
```diff
@@ -3,11 +3,12 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "2.5.0"
+version = "2.5.1"
```
devmotion, line 6: This is a new feature, not a bugfix, hence we should bump the minor version:
```suggestion
version = "2.6.0"
```

Project.toml
```diff
@@ -3,11 +3,12 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
...
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
 ConsoleProgressMonitor = "88cd18e8-d9cc-4ea6-8889-5259c0d15c8b"
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
```
devmotion, line 11: Is it actually necessary/useful to work with `DateTime`? Maybe it would be sufficient to work with `time()`?

src/metadata.jl
```diff
@@ -0,0 +1,38 @@
+using Dates
```
devmotion, line 1: Probably this should be moved to src/AbstractMCMC.jl?

src/metadata.jl
```diff
@@ -0,0 +1,38 @@
...
+
+"""
+    Metadata
```
devmotion, line 4: IMO the name `Metadata` is too generic. Maybe something like `SamplingStats` or `SamplingTimes` would be better.

src/metadata.jl
```diff
@@ -0,0 +1,38 @@
...
+
+- `start_time`: A `DateTime` indicating when sampling begun.
+- `stop_time`: A `DateTime` indicating when sampling finished.
+- `step_calls`: The number of times `step!` was called.
+- `step_time`: The total number of seconds spent inside `step!` calls.
+- `allocations`: The total number of bytes allocated by `step!`.
+"""
+mutable struct Metadata
+    start_time::DateTime
+    stop_time::Union{DateTime, Missing}
```
devmotion, line 18: Maybe just `start` and `stop`? And maybe one could just work with `Float64` here:
```suggestion
    start::Float64
    stop::Float64
```

src/metadata.jl
```diff
@@ -0,0 +1,38 @@
...
+    step_time::Float64
+    step_calls::Int64
+    allocations::Int64
```
devmotion, line 21: Better for general architectures:
```suggestion
    step_calls::Int
    allocations::Int
```

src/metadata.jl
```diff
@@ -0,0 +1,38 @@
...
+end
+
+Metadata() = Metadata(Dates.now(), missing, 0,0,0)
```
devmotion, line 24: Maybe just
```suggestion
Metadata() = Metadata(time(), Inf, 0.0, 0, 0)
```

src/metadata.jl
```diff
@@ -0,0 +1,38 @@
...
+
+function update(f, md::Metadata)
```
devmotion, line 26: I am a bit worried about all the closures that are introduced here. This puts additional pressure on the compiler, possibly leading to increased compilation time, worse type inference, and degraded performance.

src/metadata.jl
```diff
@@ -0,0 +1,38 @@
...
+    md.step_time += etime
+    md.step_calls += 1
+    md.allocations += alloc
+
+    return (sample, state)
+end
+
+function stop(md::Metadata)
+    md.stop_time = Dates.now()
+end
```
devmotion, line 38:
```suggestion
function stop!(md::Metadata)
    md.stop = time()
    return md
end
```

src/sample.jl
```diff
@@ -79,6 +79,7 @@ function mcmcsample(
     discard_initial = 0,
     thinning = 1,
     chain_type::Type=Any,
+    sampling_metadata = Metadata(),
```
devmotion, line 82: Why is a new keyword argument needed?

src/sample.jl
```diff
@@ -94,7 +95,10 @@ function mcmcsample(
         end
 
         # Obtain the initial sample and state.
-        sample, state = step(rng, model, sampler; kwargs...)
+        sample, state = update(sampling_metadata) do x
```
devmotion, line 98: As mentioned above, it might be better to avoid the closures. In any case, it seems strange to add a discarded argument?

devmotion (changes requested):
I wonder if it would be better to track statistics in downstream packages (then one could also track other stuff or function calls etc.) or with callbacks (or a combination of both). I also wonder if it is necessary to add a special struct at all - couldn't we just keep track of these things in `mcmcsample` and then possibly forward them to `bundle_samples`? This would also get rid of the closures that I would like to avoid.

If the tracking leads to worse performance (some benchmarks would be interesting), it would be good if one could disable it. One should also be aware of that the statistics will also include compilation. I am also a bit curious if it would be more relevant in general to benchmark/track the whole `sample` call with `@timed` or some more advanced approach. Is it generally useful to exclude the allocations and times arising from the other parts of the `sample` call?

cpfiffer:
> I wonder if it would be better to track statistics in downstream packages (then one could also track other stuff or function calls etc.) or with callbacks (or a combination of both). I also wonder if it is necessary to add a special struct at all - couldn't we just keep track of these things in `mcmcsample` and then possibly forward them to `bundle_samples`? This would also get rid of the closures that I would like to avoid.

I think my idea here was to initialize something a little more extensible -- we could generally start adding more tracking fields in here to provide all kinds of profiling information to users that they cannot provide themselves because we sit on top of the sampling interface.

That said, I don't like the closure much either -- adding separate counter variables at the start is maybe preferable. It's much easier to reason about the performance rather than the closure weirdness.

> If the tracking leads to worse performance (some benchmarks would be interesting), it would be good if one could disable it. One should also be aware of that the statistics will also include compilation. I am also a bit curious if it would be more relevant in general to benchmark/track the whole `sample` call with `@timed` or some more advanced approach. Is it generally useful to exclude the allocations and times arising from the other parts of the `sample` call?

Same thing above here -- I want to provide something where we could glue more stuff to the inside to track some variables users might find interesting. I think we could add a separate field to track step times and total sampling times differently -- this could help people debug IO issues (`save!!`) or whatever.

cpfiffer:
Just to add: the benefit of the struct is that someone can look at the docstring to see the available fields much more easily, so it integrates a little better with the dev experience. Not sure if that's worth the closure, though.

devmotion:
Regardless of how the data is saved and forwarded to the user, I assumed it is not necessary to save it in a struct while running `sample` and closures are not needed. To be more precise, I thought the PR could be rewritten by modifying the `mcmcsample` implementations similar to
```julia
start = time()
step_time = 0.0
allocations = 0
(state, sample), _time, _allocated, _ = @timed step(...)
step_time += _time
allocations += _allocated

....

(state, sample), _step, _allocated, _ = @timed step(...)
step_time += _time
allocations += _allocated

...

stop = time()
duration = stop - start

stats = SamplingStatistics(start, stop, step_time, Ntotal, allocations)

bundle_samples(....; stats=stats)
```

In general, I think alternatively users could benchmark the whole sample run and compute the average time and allocations per sample from this information, or downstream packages could benchmark algorithm-specific things such as specific computationally expensive operations in the sampling algorithm. Additionally, downstream packages can even benchmark `step` by defining
```julia
function step(....)
    (sample, state), time, allocation, _ = @timed _step(...) # or whatever benchmark they are interested in
    return sample, withstats(state, (; time, allocation))
end

_step(...) = ... # the actual algorithm
````
So I still wonder what the (dis)advantages over the approach in this PR are.

**Comment by cpfiffer** [comment]
Ah, okay, I like that framework better.

As to the advantages, I think this is definitely one of those things we can introduce at the interface level and give to users for free if they want it -- it's simple enough to do and is something that I think we should consider offering broadly. 

The version you put above with no closures is ridiculously cheap from a computational perspective, so I don't think it has many disadvantages.

**Comment by cpfiffer** [comment]
Alrighty, I've pulled the closures out in favor of simple incrementers. 

I've also done a little benchmarking. I ran the following line on `master` and `csp/metadata`:

```julia
@benchmark chain = sample(MyModel(), MySampler(), 10_000; chain_type = MyChain, progress=false)
```

Here it is with all the `@timed` calls for this PR:

```
BenchmarkTools.Trial:
  memory estimate:  7.17 MiB
  allocs estimate:  188978
  --------------
  minimum time:     8.009 ms (0.00% GC)
  median time:      8.613 ms (0.00% GC)
  mean time:        9.145 ms (6.23% GC)
  maximum time:     17.446 ms (24.88% GC)
  --------------
  samples:          546
  evals/sample:     1
```

And on `master`:

```
BenchmarkTools.Trial:
  memory estimate:  1.98 MiB
  allocs estimate:  78996
  --------------
  minimum time:     2.066 ms (0.00% GC)
  median time:      2.189 ms (0.00% GC)
  mean time:        2.372 ms (4.61% GC)
  maximum time:     8.861 ms (36.89% GC)
  --------------
  samples:          2100
  evals/sample:     1
```

So, about a 4x increase in times. This isn't really that much of a complicated sampling process though, so I suspect most of this is just macro overhead that is very small relative to real computational costs. To test that, I ran it on some MH code.

This PR:

```
BenchmarkTools.Trial:
  memory estimate:  149.13 MiB
  allocs estimate:  3092902
  --------------
  minimum time:     209.231 ms (7.50% GC)
  median time:      215.551 ms (8.13% GC)
  mean time:        224.546 ms (11.45% GC)
  maximum time:     385.410 ms (45.19% GC)
  --------------
  samples:          23
  evals/sample:     1
```

`master`:

```
BenchmarkTools.Trial:
  memory estimate:  90.00 MiB
  allocs estimate:  1819232
  --------------
  minimum time:     140.753 ms (7.53% GC)
  median time:      141.976 ms (7.78% GC)
  mean time:        144.517 ms (8.01% GC)
  maximum time:     185.797 ms (7.83% GC)
  --------------
  samples:          35
  evals/sample:     1
```

It's an extra 60ms on average, which is absolutely god-awful. I'll just change this over to use start, stop, and duration only. Those `@timed` calls are brutal!

devmotion:
I guess it might also help to make `SanplingStats` immutable (the values are not updated anymore) and to replace the type unions of the fields with concrete types.

cpfiffer:
Updated this to use a more broad tracker that doesn't have a bajillion macro calls inside a loop. The current benchmarks are much more in line with `master`:

```
BenchmarkTools.Trial:
  memory estimate:  90.00 MiB
  allocs estimate:  1819235
  --------------
  minimum time:     125.679 ms (6.08% GC)
  median time:      126.867 ms (6.45% GC)
  mean time:        127.549 ms (6.66% GC)
  maximum time:     143.482 ms (5.85% GC)
  --------------
  samples:          40
  evals/sample:     1
```

Project.toml
```diff
@@ -3,11 +3,12 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "2.5.0"
+version = "2.6.0"
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
 ConsoleProgressMonitor = "88cd18e8-d9cc-4ea6-8889-5259c0d15c8b"
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
```
devmotion, line 11: I guess this can be removed?
```suggestion
```

src/sample.jl
```diff
@@ -186,49 +205,69 @@ function mcmcsample(
...
-        for _ in 2:discard_initial
-            # Obtain the next sample and state.
-            sample, state = step(rng, model, sampler, state; kwargs...)
-        end
 
-        # Run callback.
-        callback === nothing || callback(rng, model, sampler, sample, 1)
+    # Start the timer
+    start = time()
+    local state
```
cpfiffer, line 211: @devmotion I added a `local` declaration here, because Julia 1.0 was not limiting scope of `state` to the `@allocated begin ...` block. Do you know if this is a big performance no-no?

cpfiffer: Nevermind, it's not actually a fix. Tests are still borked.

devmotion:
AFAICT the only problem is that on Julia 1.0 `@allocated` puts everything in a `let` block, i.e., you have to declare every variable that you want to return and use afterwards with `local`. I.e., similar to what you did with `state` you also have to add `local samples` etc.

devmotion:
A completely different design that just came to my mind: Alternatively, instead of passing a struct to `bundle_samples`, one could log the statistics with a `@debug` message. These messages can contain keys with values and they are cheap since their computation and creating the message are skipped completely if debugging is not enabled. Downstream packages would have to handle these messages if they want to use the stats (e.g., with LoggingExtras) which might be a bit more inconvenient than passing it to `bundle_samples` - on the other hand, maybe it could be useful if not every implementation of `bundle_samples` has to implement how to deal with the stats in this way.

cpfiffer:
I dunno, seems very low-cost to include -- I wouldn't want someone to catch rando log messages. Seems messy to me.

Maybe some context is important here. I wanted to add this because I've had some requests for things like ESS/minute, which requires time information, and I figured anyone who used AbstractMCMC might also want time information at some level.

cpfiffer:
I pulled out the allocation tracker, so this is is now a much simpler PR. All it tracks is start time, stop time, and duration. Later on we can consider more profiling tools for interested users, but at the moment this is a nice, simple, low-cost addition.

devmotion (commented):
👍 Have you benchmarked the latest version? I agree with you, I wouldn't expect any performance regressions though. Maybe it would be good to mention the new keyword argument to `bundle_samples` in the README?

cpfiffer:
The benchmarks are basically equivalent to the current master:

```
BenchmarkTools.Trial:
  memory estimate:  90.00 MiB
  allocs estimate:  1819233
  --------------
  minimum time:     131.155 ms (7.98% GC)
  median time:      132.929 ms (8.19% GC)
  mean time:        133.656 ms (8.42% GC)
  maximum time:     151.900 ms (7.09% GC)
  --------------
  samples:          38
  evals/sample:     1
```

I've added a little bit to the README as well.

devmotion (approved):
LGTM 👍

PR #66, opened 2021-04-10
fix callback signature in convergence mcmcsample

mileslucas:
#56 changed the callback signature but did not update it for the convergence sampling `mcmcsample`.


devmotion:
Can you update the version number (it should be a bugfix release) and add some tests?

cpfiffer (approved):

PR #68, opened 2021-04-10
add `state` to isdone signature

### LINKED ISSUE #67: Update convergence sampling signature to use `state`
https://github.com/TuringLang/AbstractMCMC.jl/blob/d961513b0ad3dc53d414da6581b332b66c4d0ef3/src/sample.jl#L209

I think this should be updated to be
```julia
isdone(rng, model, sampler, samples, state, i; progress=progress, kwargs...)
```
to coincide with the immutable samplers (#42, #56), which would be breaking (unfortunately).

Thoughts on this? I'm happy to PR, just let me know if this has downstream effects that I'll need to address, too.

mileslucas:
Breaking change to `isdone` signature

fixes #67

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "4.0.0"
+version = "3.0.2"
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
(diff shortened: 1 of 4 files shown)


Project.toml
```diff
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "3.0.1"
+version = "4.0.0"
```
cpfiffer, line 6:
```suggestion
version = "3.0.2"
```

I think this should be considered part of the 3.0 break, since it's all the same general change.

mileslucas: Sounds good to me

PR #71, opened 2021-05-09
Add MCMCSerial

cpfiffer:
I often run into the case where I want to get multiple chains but don't want to go through the hassle of making things threadsafe or available to each process. The common way to do this currently is something like

```julia
chains = map(x -> sample(model, spl, n), 1:10) # sample ten chains
chain = reduce(chainscat, chains)
```

Which is fine, but it is odd that we don't have a convenient way of sampling multiple chains in a basic way. 

I added `MCMCSerial` as a parallelization subtype, so you can now simply call

```julia
chain = sample(model, spl, MCMCSerial(), n, 10)
```

which will sample each subchain in turn and concatenate them. Tests are included.

I'm also open to changing the call here --  it seems to me that this should be the default behavior of something like

```julia
sample(model, spl, n, n_chains)
```

where no parallization method is specified. I haven't implemented the signature above but I can if people think it's a good idea.

### DIFF
--- a/src/AbstractMCMC.jl
+++ b/src/AbstractMCMC.jl
@@ -55,9 +55,6 @@ in parallel.
 """
 abstract type AbstractMCMCEnsemble end
 
-# Deprecate the old name AbstractMCMCParallel in favor of AbstractMCMCEnsemble
-Base.@deprecate_binding AbstractMCMCParallel AbstractMCMCEnsemble false
-
 """
     MCMCThreads
 
@@ -88,5 +85,6 @@ include("interface.jl")
 include("sample.jl")
 include("stepper.jl")
 include("transducer.jl")
+include("deprecations.jl")
 
 end # module AbstractMCMC
--- a/src/AbstractMCMC.jl
+++ b/src/AbstractMCMC.jl
@@ -48,12 +48,15 @@ An `AbstractModel` represents a generic model type that can be used to perform i
 abstract type AbstractModel end
 
 """
-    AbstractMCMCParallel
+    AbstractMCMCEnsemble
 
-An `AbstractMCMCParallel` algorithm represents a specific algorithm for sampling MCMC chains
+An `AbstractMCMCEnsemble` algorithm represents a specific algorithm for sampling MCMC chains
 in parallel.
 """
-abstract type AbstractMCMCParallel end
+abstract type AbstractMCMCEnsemble end
+
+# Deprecate the old name AbstractMCMCParallel in favor of AbstractMCMCEnsemble
+Base.@deprecate_binding AbstractMCMCParallel AbstractMCMCEnsemble false
 
 """
     MCMCThreads
@@ -61,7 +64,7 @@ abstract type AbstractMCMCParallel end
 The `MCMCThreads` algorithm allows users to sample MCMC chains in parallel using multiple
 threads.
 """
-struct MCMCThreads <: AbstractMCMCParallel end
+struct MCMCThreads <: AbstractMCMCEnsemble end
 
 """
     MCMCDistributed
@@ -69,7 +72,7 @@ struct MCMCThreads <: AbstractMCMCParallel end
 The `MCMCDistributed` algorithm allows users to sample MCMC chains in parallel using multiple
 processes.
 """
-struct MCMCDistributed <: AbstractMCMCParallel end
+struct MCMCDistributed <: AbstractMCMCEnsemble end
 
 
 """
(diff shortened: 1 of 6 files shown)


src/sample.jl
```diff
@@ -444,5 +444,35 @@ function mcmcsample(
...
+    # Set up a chains vector.
+    chains = Vector{Any}(undef, nchains)
+
+    # Sample each chain
+    for i in 1:nchains
+        # Sample a chain and save it to the vector.
+        chains[i] = StatsBase.sample(rng, model, sampler, N; 
+                                     progressname = string(progressname, " (Chain $i of $nchains)"),
+                                     kwargs...)
+    end
```
devmotion, line 471: Could one use `map` to avoid having to preallocate `chains`?

src/AbstractMCMC.jl
```diff
@@ -58,19 +58,27 @@ abstract type AbstractMCMCParallel end
...
 """
 struct MCMCDistributed <: AbstractMCMCParallel end
 
+
+"""
+    MCMCSerial
+
+The `MCMCSerial` algorithm allows users to sample serially, with no thread or process parallelism.
+"""
+struct MCMCSerial <: AbstractMCMCParallel end
```
devmotion, line 80: The supertype is a bit weird...

devmotion (commented):
I think it's definitely a good idea to add something like `MCMCSerial` :+1: I had the same thoughts for a while, the current approach feels quite cumbersome.

> it seems to me that this should be the default behavior of something like
> ```julia
> sample(model, spl, n, n_chains)
> ```
> where no parallization method is specified.

I'm not completely sure, I thought that maybe rather `MCMCThreads` should be the default - e.g., it is the default context in JuliaFolds and also in other packages such as DataFrames and CSV multithreading is the default option.

src/AbstractMCMC.jl
cpfiffer: Agreed, maybe change it to `AbstractMCMCMultichain` or something less parallel-specific?

src/sample.jl
cpfiffer: Yeah, done.

cpfiffer:
> I'm not completely sure, I thought that maybe rather `MCMCThreads` should be the default - e.g., it is the default context in JuliaFolds and also in other packages such as DataFrames and CSV multithreading is the default option.

Okay, let's leave it be for now -- I don't think we should use threads as the default because we have so little control over the downstream operation. Lots of things are not threadsafe still and MCMC models can have weird things in them that don't lend themselves to thread parallelization well. We'll stick with the "specify a method" thing for now, and then we can gauge later who is using what as the default method.

src/AbstractMCMC.jl
devmotion: I don't have a good suggestion, maybe just `AbstractMCMCMulti` or something like `AbstractMCMCEnsemble`, similar to the `EnsembleProblem`s etc. in SciML (they cover also distributed, multi-threaded, and serial computations).

In any case, one could make the renaming non-breaking (I think) if one deprecates `AbstractMCMCParallel` with [`Base.@deprecate_binding`](https://github.com/JuliaLang/julia/blob/c3a31804a88beb4b05c8fd4b0981e76d14058311/base/deprecated.jl#L137):
```julia
Base.@deprecate_binding AbstractMCMCParallel AbstractMCMCEnsemble false
```

src/sample.jl
```diff
@@ -444,5 +444,31 @@ function mcmcsample(
...
+    kwargs...
+)
+    # Check if the number of chains is larger than the number of samples
+    if nchains > N
+        @warn "Number of chains ($nchains) is greater than number of samples per chain ($N)"
+    end
+
+    # Sample the chains.
+    chains = map(
+        i -> StatsBase.sample(rng, model, sampler, N; progressname = string(progressname, " (Chain $i of $nchains)"),
```
devmotion, line 464: One could avoid interpolation since you use `string` anyway:
```suggestion
        i -> StatsBase.sample(rng, model, sampler, N; progressname = string(progressname, " (Chain ", i, " of ", nchains, ")"),
```

src/AbstractMCMC.jl
cpfiffer: Okay, I like that. I've switched to AbstractMCMCEnsemble.

src/AbstractMCMC.jl
```diff
@@ -48,28 +48,39 @@ An `AbstractModel` represents a generic model type that can be used to perform i
...
 
-An `AbstractMCMCParallel` algorithm represents a specific algorithm for sampling MCMC chains
+An `AbstractMCMCEnsemble` algorithm represents a specific algorithm for sampling MCMC chains
 in parallel.
 """
-abstract type AbstractMCMCParallel end
+abstract type AbstractMCMCEnsemble end
+
+# Deprecate the old name AbstractMCMCParallel in favor of AbstractMCMCEnsemble
+Base.@deprecate_binding AbstractMCMCParallel AbstractMCMCEnsemble false
```
devmotion, line 59: Maybe collect the deprecations at the end of the file or even a separate file? Then it would be easier to spot them and remove them in the next breaking release.

devmotion (approved):
Looks good, just one minor suggestion which you can ignore if you don't agree.

PR #76, opened 2021-06-09
Build stable documentation

devmotion:
I just noticed that no docs exist for tagged releases. The problem is that TagBot is not allowed to trigger other actions.


devmotion:
I used the opportunity to update CompatHelper as well, it uses the Documenter key as well now.

cpfiffer (approved):

Issue #82, opened 2021-08-31
More useful logging when sampling multiple chains


sethaxen:
If I sample multiple chains, the progress bar shown in the terminal only updates when a chain finishes sampling. This perhaps makes sense if I am quickly sampling many chains in parallel, but this is essentially useless for the common case of a user sampling 2-8 chains of medium to long runtimes in parallel, since as a user I don't have an estimate if the chains will finish sampling in 4 minutes or 4 days.

Currently one works around this by sampling a single chain "long enough" to guess how long it will take to finish, killing the sampler, and then restarting with multiple chains, but it would be better if multiple progress bars were shown, one for each chain, perhaps with a global progress bar showing how many chains have finished. On Slack a while ago, someone proposed that if a user sampled more chains than progress bars would fit in the terminal, then it might be better to use something like the animation Julia uses when compiling packages.

Are the limitations to doing something like this in this package or one of its dependencies?

devmotion:
This is a longstanding issue and should be improved. The main reason for why currently not the progress of individual samples is shown is that it is a bit challenging to do it right: it is expected that users only have to implement the `sample` method for a single chain and then the multi-chain methods work automatically. Ie, we always want to call the single chain method in the multi chain methods.

Some explanation of the progress logging in general (copied from Slack):

> I assume that Pluto uses a special logger or handles logging messages in a different way than what is done in the REPL. Maybe logging messages are dropped completely? We use the frontend agnostic ProgressLogging.jl for progress logging (also used eg by SciML). It uses the logging framework in Julia and works in the REPL (with TerminalLoggers), Juno, VSCode (both use ProgressLogging for progress bars anyway) and IJulia. We also make sure that there's a logger that can handle the messages, and add a local logger for the progress logs otherwise which is optimized for the environment that code is run in (IIRC for IJulia and Windows we use ConsoleProgressMonitor instead of TerminalLogger).

devmotion:
Related: https://github.com/JuliaLogging/ProgressLogging.jl/issues/27, https://github.com/JuliaLogging/ProgressLogging.jl/issues/33, https://github.com/JuliaLogging/ProgressLogging.jl/issues/37, https://github.com/JuliaLogging/ProgressLogging.jl/issues/38

kaandocal:
From what I understand the parallel `mcmcsample` implementations call the serial `mcmcsample` in each thread/process, so maybe one could hook each sampler to a callback which updates the corresponding progress bar...

[...]

PR #83, opened 2021-09-01
Testing new progress bars

kaandocal:
This is a proposed fix for #82 (definitely WIP). I created variants of some ProgressLogging macros that can be used to add children to the current progress bar, see src/logging.jl; these probably belong elsewhere.

### DIFF
--- a/src/logging.jl
+++ b/src/logging.jl
@@ -43,3 +43,134 @@ function progresslogger()
         return TerminalLoggers.TerminalLogger()
     end
 end
+
+##
+
+using ProgressLogging: _asprogress, ProgressString, ProgressLevel, uuid4
+using Base.Meta: isexpr
+
+const _id_var_children = gensym(:progress_id_children)
+const _name_var_children = gensym(:progress_name_children)
+
+macro progressid_child(N)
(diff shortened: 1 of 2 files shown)


devmotion:
> EDIT: Any opinions about ProgressMeter.jl?

We replaced it with ProgressLogging 😛 The main advantage of ProgressLogging is that it uses the Julia logging system and works with different frontends (eg TerminalLogger, ConsoleProgressMonitor, Juno, VSCode). However, there are still things missing/unpolished in the ProgressLogging API, as the references in the linked issue show, and some things are not supported by the loggers.

src/logging.jl
```diff
@@ -43,3 +43,134 @@ function progresslogger()
         return TerminalLoggers.TerminalLogger()
     end
 end
+
+##
+
+using ProgressLogging: _asprogress, ProgressString, ProgressLevel, uuid4
```
devmotion, line 49: IMO the use of these internals indicates that these extensions should be discussed and potentially be added in ProgressLogging but not in AbstractMCMC.

kaandocal: No doubt, I will test these extensions and  may open a PR in ProgressLogging later, this is just a quick hack

Issue #85, opened 2021-10-17
Improving interaction between different implementations of the interface


torfjelde:
How do you feel about adding something like:
```julia
"""
    state_from_transiton(state, transition_prev[, state_prev])

Return new instance of `state` using information from `transition_prev` and, optionally, `state_prev`.

Defaults to `setparameters!!(state, parameters(transition_prev))`.
"""
function state_from_transition(state, transition_prev, state_prev)
    return state_from_transition(state, transition_prev)
end

function state_from_transition(state, transition)
    return setparameters!!(state, parameters(transition))
end

"""
    setparameters!!(state, parameters)

Return new instance of `state` with parameters set to `parameters`.
"""
setparameters!!

"""
    parameters(transition)

Return parameters in `transition`.
"""
parameters
```
to make it easier for samplers to interact across packages? Then you just need to implement `state_from_transition` for the different types to get cross-package compat. 

Another issue is also the `model` argument which often is specific to a particular sampler implementation. We've previous spoken about generalizing this so that we don't have a bunch of these lying around, e.g. `AdvancedMH.DensityModel` and `AdvancedHMC.DifferentiableDensityModel`, but we should also maybe add a function `getmodel(model, sampler, state)` or something too,  which is `identity` by default but allows one to provide a model-type which encodes a bunch of different models for specific samplers, e.g. in the case of a `MixtureSampler` you might have a `MixtureState` which, among other things, holds the current sampler-index, and a `ManyModels` which simply wraps a collection of models corresponding to each of the components/samplers in the `MixtureSampler`:

```julia
getmodel(model::ManyModels, sampler::MixtureSampler, state::MixtureState) = model[state.current_index]
```

I've been running into quite a few scenarios recently where I'd love to have something like this, e.g. wanting to implement `MixtureSampler` and `CompositionSampler`.

torfjelde:
@devmotion @cpfiffer @yebai thoughts?

torfjelde:
Thinking about this again due to work on Gibbs sampler for Turing.jl.

I think we need the following from a `AbstractSampler` to do "most" interesting "meta"-samplers:
1. Ability to get and set parameters and logprob.
2. Ability to go from the state of one sampler, say, `SamplerA`, to the state of another sampler, say, `SamplerB`.
2. Ability to re-evaluate the logprob as the sampler would.
   - Crucial part is the "as the sampler would", e.g. might need to recompute gradients too.


In short, we need something like:

``` julia
[...]

PR #86, opened 2021-10-21
Add  `getparameters` and `setparameters!!`

torfjelde:
See #85.

Currently, almost every package implementing a `AbstractSampler` comes with their custom state and transition types. Since there currently are no "interface" or "guidelines" about what the types of `state` and `transition` should implement, it means that interaction across different samplers, e.g. composing two samplers from different packages, is annoying.

This PR introduces some methods which aims to alleviate this issue. In particular it sampler-implementations to also implement
- `parameters(transition)` for the transition-type which returns the parameters in the sample, where by "parameters" I mean the realizations of the random variables we're targeting present in the `transition`, and
- `setparameters!!` for the state-type which updates the current state of the Markov chain, thus likely (this might depend on the sampler) conditioning the sampler/kernel on the provided parameters.

*Why `parameters` for `transition` but `setparameters!!` for `state`?* A `transition` should (AFAIK) always contain a realization of the random variables we're sampling while the `state` might not, e.g. MH with a `MvNormal(zeros(n), I)` as proposal distribution (not dependent on current position/state/parameters). Of course there might be more properties one would like to use when converting from `PackageA.Transition` to `PackageB.State`, but this seems like the lowest common denominator that we can come up with. Maaaybe we should have the same for `logdensity`, i.e. `logdensity(transition)` and `setlogdensity!!(state)`, but this might not always be available nor correspond directly to the same thing for both samplers. Might be worth including though. They could also obviously be implemented for _both_ if the implementer so chooses.

*But* for the cases where you want to make use of more than just the parameters (and potentially logdensity), one can overload the `state_from_transition` directly (which is what should be used by packages making these "meta-samplers", e.g. composition), e.g. `AdvancedHMC` might want to implement `state_from_transition(::AdvancedHMC.HMCState, ::AdvancedHMC.Transition)` to not only update the parameters in the `state` but the entire `PhasePoint` from `AdvnacedHMC.Transition`, etc.

Thoughts?


src/AbstractMCMC.jl
```diff
@@ -79,6 +79,36 @@ The `MCMCSerial` algorithm allows users to sample serially, with no thread or pr
 """
 struct MCMCSerial <: AbstractMCMCEnsemble end
 
+"""
+    state_from_transiton(state, transition_prev[, state_prev])
```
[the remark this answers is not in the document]
torfjelde: Yo, whadda yah guys think?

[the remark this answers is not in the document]
cpfiffer: Oh, sorry, my bad. Yeah your proposed name change is clearer, let's do that.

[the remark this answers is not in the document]
torfjelde: Or maybe just `update!!`? Given that we might also want to pass in `state_prev`?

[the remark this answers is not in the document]
devmotion: Yeah, since we don't want to export any of these functions anyway I'm fine with `update!!`. Is a bit nicer than `update_from_transition!!` IMO, and as you say we might not only update based on the transition.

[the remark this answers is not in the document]
torfjelde: Cool, I'll make that change then :+1:

torfjelde:
Do you have ideas for alternatives to the use of `parameters` @devmotion and @cpfiffer ? As I said in my previous comment, I think "samples" is kind of confusing too. Really what we want is "set the realizations of the random variables" but I think "realizations" is too unfamiliar for people.

I think I'm leaning towards `values` and `setvalues!!`.

cpfiffer:
I dunno I feel like I wouldn't mind using `realize` or `realize!!`. Short and sweet, and I don't think it's _that_ out of the way. If that doesn't seem right, I don't mind using `values` (though I will note that it's a base function override, which I don't care about overmuch).

torfjelde:
I've added some docs + motivating example: https://turinglang.github.io/AbstractMCMC.jl/previews/PR86/api/#Interacting-with-states-of-samplers

(Btw, I love the fact that I can preview docs in the PRs; I'm guessing this is your doing @devmotion ? :heart: )

cpfiffer (approved):
Mhm yep I love it. Could you move us up a major version?

torfjelde:
I think the only thing that's left is the discussion of `parameters` vs. `realizations` vs. `values`

cpfiffer:
I prefer `realizations` but `values` is more Julian -- let's go with that.

torfjelde:
On additional thing we might want to consider adding: `logdensity` and `setlogdensity!!`, or something along those lines. This is because:
1. Almost _all_ samplers will have a cached value of the logdensity (up-to some constant) of their target in their state.
2. Some samplers will make use of said cached value, e.g. AdvancedMH uses this in its accept/reject computation.

Of course, this can be dealt with on a case-by-case basis by simply defining `updatestate!!` explicitly, _but_ this will be unnecessary in _most_ cases if we have something along the lines of `logdensity` and `setlogdensity!!`.

torfjelde:
Also, @devmotion you mentioned tests. Do you have anything particular in mind?

I have a test for the `MixtureSampler` outlined in the docs using AdvancedMH.jl, but that seems a bit "too heavy". At the same time, doing something like:
```julia
struct TestState
    params
    logp
end

setvalues!!(s::TestState) = ...
```
etc. seems quite useless :shrug: 

We could potentially take the test I have for the `MixtureSampler` and add a super-simple implementation of `RWMH` to the test-suite of AbstractMCMC? Not sure I'm too big a fan of this either though, as it'll be more like an example use-case rather than testing the functionality provided by `updatestate!!` etc.

torfjelde:
> On additional thing we might want to consider adding: `logdensity` and `setlogdensity!!`, or something along those lines. This is because:
> 
> 1. Almost _all_ samplers will have a cached value of the logdensity (up-to some constant) of their target in their state.
> 2. Some samplers will make use of said cached value, e.g. AdvancedMH uses this in its accept/reject computation.
> 
> Of course, this can be dealt with on a case-by-case basis by simply defining `updatestate!!` explicitly, _but_ this will be unnecessary in _most_ cases if we have something along the lines of `logdensity` and `setlogdensity!!`.
> 
> What do you say @devmotion @cpfiffer ? Add or not add, that is the question.

Yo boys @devmotion @cpfiffer! You got any thoughts on the above? AFAIK this is the only thing that's holding this PR back.

src/AbstractMCMC.jl
```diff
@@ -79,6 +79,34 @@ The `MCMCSerial` algorithm allows users to sample serially, with no thread or pr
...
+values. Otherwise a new `state` object with the new `values` is returned.
+"""
+function setvalues!! end
+
+@doc """
+    values(transition)
+
+Return values in `transition`.
+"""
+Base.values
```
devmotion, line 107: Do we really want to use `Base.values` here? I think adding this documentation should be considered type piracy and would only be allowed if we define it only for our own types as arguments (which we can't do here).

Unrelated, why did you use `@doc`?

cpfiffer:
> Yo boys @devmotion @cpfiffer! You got any thoughts on the above? AFAIK this is the only thing that's holding this PR back.

Woops, sorry, missed this one. Full support from me on this one!

devmotion:
Ah I also missed this question. I think it would probably be useful but I would leave it for a separate PR. There were quite many discussions about `logdensity` functions recently and there was a push for unifying them throughout the Julia ecosystem. DensityInterface.jl defines a common interface for log density functions and their evaluations now that is used by e.g. Distributions, AbstractPPL, and (AFAIK) in the future also MeasureTheory. Hence IMO we should be a bit careful and not introduce a different `logdensity` function here, now that we unified them, but also make sure that it corresponds to the DensityInterface assumptions (if we use its interface).

src/AbstractMCMC.jl
torfjelde: > Do we really want to use Base.values here? I think adding this documentation should be considered type piracy and would only be allowed if we define it only for our own types as arguments (which we can't do here).

Hmm, I do agree with this. I'm coming around to `values` not being the best idea.

> Unrelated, why did you use @doc?

IIRC you need `@doc` here to be able to add the docstring? I.e. if you remove the `@doc` it won't be added to the docstring of `Base.values`.

torfjelde:
Yeah I was actually looking at the newly introduced `logdensity` stuff today, so I agree that maybe this should be done in a separate PR :+1:

torfjelde:
I renamed `values` and `setvalues!!` to `realize` and `realize!!` as mentioned earlier.

Even if we're not completely happy with these names, we can instead just change these at a later stage.

torfjelde:
> Ah I also missed this question. I think it would probably be useful but I would leave it for a separate PR. There were quite many discussions about logdensity functions recently and there was a push for unifying them throughout the Julia ecosystem. DensityInterface.jl defines a common interface for log density functions and their evaluations now that is used by e.g. Distributions, AbstractPPL, and (AFAIK) in the future also MeasureTheory. Hence IMO we should be a bit careful and not introduce a different logdensity function here, now that we unified them, but also make sure that it corresponds to the DensityInterface assumptions (if we use its interface).

After looking into DensityInterface.jl I don't think there's anything that should be relevant for AbstractMCMC's functionality for setting and extracting log-probabilities _from states/transitions_. For the `AbstractModel`, etc., then it's def relevant though!

So maybe we should just introduce `logprob` or `logtargetprob` or maybe even just `logtarget`, and its corresponding `set..!!`? I'm realizing more and more that it's necessary to be able to transfer the log target-prob to the next state if we want to do something interesting. In fact, I think `update!!` state should optionally be able to take the `AbstractModel` that we're sampling from too so that it can re-evaluate the current state if need be.

src/AbstractMCMC.jl
```diff
@@ -79,6 +79,36 @@ The `MCMCSerial` algorithm allows users to sample serially, with no thread or pr
...
+"""
+function realize!! end
+
+"""
+    realize(transition)
+
+Return the realization of the random variables present in `transition`.
+"""
+function realize end
+
```
sunxd3, line 110:
```suggestion

"""
    getparams(state[; kwargs...])

Retrieve the values of parameters from the sampler's `state` as a `Vector{<:Real}`.
"""
function getparams end

"""
    setparams!!(state, params)

Set the values of parameters in the sampler's `state` from a `Vector{<:Real}`. This function should
follow the `BangBang` interface: mutate `state` in-place if possible and return the
mutated `state`. Otherwise, it should return a new `state` containing the updated parameters.

Although not enforced, it should hold that `setparams!!(state, getparams(state)) == state`. In another
word, the sampler should implement a consistent transformation between its internal representation
and the vector representation of the parameter values.
"""
function setparams!! end
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
 AbstractMCMC.chainscat
 AbstractMCMC.chainsstack
 ```
+
+## Interacting with states of samplers
+
+To make it a bit easier to interact with some arbitrary sampler state, we encourage implementations of `AbstractSampler` to implement the following methods:
+```@docs
+AbstractMCMC.realize
+AbstractMCMC.realize!!
```
sunxd3, line 116:
```suggestion
AbstractMCMC.getparams
AbstractMCMC.setparams!!
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+```
+and optionally
+```@docs
+AbstractMCMC.updatestate!!(state, transition, state_prev)
+```
```
sunxd3, line 121:
```suggestion
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+These methods can also be useful for implementing samplers which wraps some inner samplers, e.g. a mixture of samplers.
+
+### Example: `MixtureSampler`
+
+In a `MixtureSampler` we need two things:
+- `components`: collection of samplers.
+- `weights`: collection of weights representing the probability of chosing the corresponding sampler.
```
sunxd3, line 128:
```suggestion
- `weights`: collection of weights representing the probability of choosing the corresponding sampler.
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+- `index`: the index of the sampler used in this `step`.
+- `transition`: the transition resulting from this `step`.
+- `states`: the current states of _all_ the components.
+Two aspects of this might seem a bit strange:
+1. We need to keep track of the states of _all_ components rather than just the state for the sampler we used previously.
+2. We need to put the `transition` from the `step` into the state.
+
+The reason for (1) is that lots of samplers keep track of more than just the previous realizations of the variables, e.g. in `AdvancedHMC.jl` we keep track of the momentum used, the metric used, etc.
+
+For (2) the reason is similar: some samplers might keep track of the variables _in the state_ differently, e.g. you might have a sampler which is _independent_ of the current realizations and the state is simply `nothing`.
```
sunxd3, line 147:
```suggestion
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+```
+The `step` for a `MixtureSampler` is defined by the following generative process
+```math
+\begin{aligned}
+i &\sim \mathrm{Categorical}(w_1, \dots, w_k) \\
+X_t &\sim \mathcal{K}_i(\cdot \mid X_{t - 1})
+\end{aligned}
+```
+where ``\mathcal{K}_i`` denotes the i-th kernel/sampler, and ``w_i`` denotes the weight/probability of choosing the i-th sampler.
+[`AbstractMCMC.updatestate!!`](@ref) comes into play in defining/computing ``\mathcal{K}_i(\cdot \mid X_{t - 1})`` since ``X_{t - 1}`` could be coming from a different sampler.
```
sunxd3, line 165:
```suggestion
[`AbstractMCMC.getparams`](@ref) and [`AbstractMCMC.setparams!!`](@ref) comes into play in defining/computing ``\mathcal{K}_i(\cdot \mid X_{t - 1})`` since ``X_{t - 1}`` could be coming from a different sampler.
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+[`AbstractMCMC.updatestate!!`](@ref) comes into play in defining/computing ``\mathcal{K}_i(\cdot \mid X_{t - 1})`` since ``X_{t - 1}`` could be coming from a different sampler. 
+
+If we let `state` be the current `MixtureState`, `i` the current component, and `i_prev` is the previous component we sampled from, then this translates into the following piece of code:
+
+```julia
+# Update the corresponding state, i.e. `state.states[i]`, using
+# the state and transition from the previous iteration.
+state_current = AbstractMCMC.updatestate!!(
+    state.states[i], state.states[i_prev], state.transition
+)
```
sunxd3, line 174:
```suggestion
state_current = AbstractMCMC.setparams!!(
    state.states[i], 
    AbstractMCMC.getparams(state.states[i_prev]),
)
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+```julia
+struct MixtureSampler{W,C} <: AbstractMCMC.AbstractSampler
+    components::C
+    weights::W
+end
+```
+
+To implement the state, we need to keep track of a couple of things:
+- `index`: the index of the sampler used in this `step`.
+- `transition`: the transition resulting from this `step`.
```
sunxd3, line 139:
```suggestion
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+- `states`: the current states of _all_ the components.
+Two aspects of this might seem a bit strange:
+1. We need to keep track of the states of _all_ components rather than just the state for the sampler we used previously.
+2. We need to put the `transition` from the `step` into the state.
+
+The reason for (1) is that lots of samplers keep track of more than just the previous realizations of the variables, e.g. in `AdvancedHMC.jl` we keep track of the momentum used, the metric used, etc.
```
sunxd3, line 145:
```suggestion
We need to keep track of the states of _all_ components rather than just the state for the sampler we used previously.
The reason is that lots of samplers keep track of more than just the previous realizations of the variables, e.g. in `AdvancedHMC.jl` we keep track of the momentum used, the metric used, etc.
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+
+For (2) the reason is similar: some samplers might keep track of the variables _in the state_ differently, e.g. you might have a sampler which is _independent_ of the current realizations and the state is simply `nothing`. 
+
+Hence, we need the `transition`, which should always contain the realizations, to make sure we can resume from the same point in the space in the next `step`.
```
sunxd3, line 149:
```suggestion
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+```julia
+struct MixtureState{T,S}
+    index::Int
+    transition::T
```
sunxd3, line 153:
```suggestion
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+1. We need to keep track of the states of _all_ components rather than just the state for the sampler we used previously.
+2. We need to put the `transition` from the `step` into the state.
+
+The reason for (1) is that lots of samplers keep track of more than just the previous realizations of the variables, e.g. in `AdvancedHMC.jl` we keep track of the momentum used, the metric used, etc.
+
+For (2) the reason is similar: some samplers might keep track of the variables _in the state_ differently, e.g. you might have a sampler which is _independent_ of the current realizations and the state is simply `nothing`. 
+
+Hence, we need the `transition`, which should always contain the realizations, to make sure we can resume from the same point in the space in the next `step`.
+```julia
+struct MixtureState{T,S}
```
sunxd3, line 151:
```suggestion
struct MixtureState{S}
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+    # Sample the component to use in this `step`.
+    i = rand(Categorical(sampler.weights))
+    sampler_current = sampler.components[i]
+
+    # Update the corresponding state, i.e. `state.states[i]`, using
+    # the state and transition from the previous iteration.
+    i_prev = state.index
+    state_current = AbstractMCMC.updatestate!!(
+        model, state.states[i], state.states[i_prev], state.transition
+    )
```
sunxd3, line 196:
```suggestion
    state_current = AbstractMCMC.setparams!!(  
        state.states[i], 
        AbstractMCMC.getparams(state.states[i_prev]),  
    )
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+            # Replace the i-th state with the new one.
+            state_current
+        else
+            # Otherwise we just carry over the previous ones.
+            state.states[j]
+        end
+    end
+
+    # Create the new `MixtureState`.
+    state_new = MixtureState(i, transition, states_new)
```
sunxd3, line 220:
```suggestion
    state_new = MixtureState(i, states_new)
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+    end
+
+    # Sample the component to use this `step`.
+    i = rand(Categorical(sampler.weights))
+    # Extract the corresponding transition.
+    transition = first(transitions_and_states[i])
+    # Extract states.
+    states = map(last, transitions_and_states)
+    # Create new `MixtureState`.
+    state = MixtureState(i, transition, states)
```
sunxd3, line 242:
```suggestion
    state = MixtureState(i, states)
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+
+    return transition, state
+end
+```
+
+Suppose we then wanted to use this with some of the packages which implements AbstractMCMC.jl's interface, e.g. [`AdvancedMH.jl`](https://github.com/TuringLang/AdvancedMH.jl), then we'd simply have to implement `realize` and `realize!!`:
```
sunxd3, line 248:
```suggestion
Suppose we then wanted to use this with some of the packages which implements AbstractMCMC.jl's interface, e.g. [`AdvancedMH.jl`](https://github.com/TuringLang/AdvancedMH.jl), then we'd simply have to implement `getparams` and `setparams!!`.
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+
+```julia
+function AbstractMCMC.updatestate!!(model, ::AdvancedMH.Transition, state_prev::AdvancedMH.Transition)
+    # Let's `deepcopy` just to be certain.
+    return deepcopy(state_prev)
+end
+```
```
sunxd3, line 255:
```suggestion
```

docs/src/api.md
```diff
@@ -107,3 +107,181 @@ For chains of this type, AbstractMCMC defines the following two methods.
...
+
+```julia
+# Take a `step` for this sampler using the updated state.
+transition, state_current = AbstractMCMC.step(
+    rng, model[i], sampler_current, state_current;
+    kwargs...
+)
+```
+
+This issue should eventually disappear as the community moves towards a unified approach to implement `AbstractMCMC.AbstractModel`.
```
sunxd3, line 287:
```suggestion
```

sunxd3 (changes requested):
Let's make a final push to get this merged!

Apologize for the amount of comments. My proposal here is, in spirit, the same as what the PR already has. The differences are:

1. not defining `updatestates!!`: sampler packages can define there own internal functions
2. renaming `realize` and `realize!!` to `getparams` and `setparams!!`: there are no perfect names here, these are the names we are currently using in `Turing.jl`, they are   self-explanatory enough
3. no longer taking `model` as argument, also force returning and receiving `Vector`: this makes the proposed interface less flexible, but I think they are sufficient, when needs arise, we can always add version with `model` arguements

docs/src/api.md
torfjelde: Why are we removing this?

sunxd3: My bad, I prepared to write some explanation but forgot. 

The deleted part was to explain 
```
2. We need to put the `transition` from the `step` into the state.
```

because the `transition` is no longer needed (we effectively force samplers to store params in state), I felt we don't need this line anymore

torfjelde: Ah gotcha :+1: Thanks @sunxd3 !

sunxd3:
@hong @devmotion can you give a quick look?
If no objections, I may merge this by the end of today (Oct 11).

sunxd3:
minor version bumped for added feature, merging now.

PR #87, opened 2021-11-07
Make subrngs reusable in parallel mcmcsample

mohamed82008:
In this PR, I attempt to expose the sub random number generators and make them reusable between calls to `mcmcsample`. This is helpful to maintain the state of the sampling. Unfortunately, it seems that I have broken the reproducibility tests. I am not exactly sure how my change here broke things. Could use extra pairs of eyes.

### DIFF
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -284,8 +284,23 @@ function mcmcsample(
     nchains::Integer;
     progress = PROGRESS[],
     progressname = "Sampling ($(min(nchains, Threads.nthreads())) threads)",
+    rngs::Union{Nothing, AbstractVector{<:Random.AbstractRNG}} = nothing,
     kwargs...
 )
+    interval = 1:min(nchains, Threads.nthreads())
+    if rngs === nothing
+        # Copy the random number generator, model, and sample for each thread
+        # NOTE: As of May 17, 2020, this relies on Julia's thread scheduling functionality
+        #       that distributes a for loop into equal-sized blocks and allocates them
+        #       to each thread. If this changes, we may need to rethink things here.
+        rngs = [deepcopy(rng) for _ in interval]
+        # Create a seed for each chain using the provided random number generator.
+        seeds = rand(rng, UInt, nchains)
+        Random.seed!.(rngs, seeds)
+    else
+        @assert length(rngs) == interval
+    end
+
     # Check if actually multiple threads are used.
     if Threads.nthreads() == 1
         @warn "Only a single thread available: MCMC chains are not sampled in parallel"
@@ -296,17 +311,8 @@ function mcmcsample(
         @warn "Number of chains ($nchains) is greater than number of samples per chain ($N)"
     end
 
-    # Copy the random number generator, model, and sample for each thread
-    # NOTE: As of May 17, 2020, this relies on Julia's thread scheduling functionality
-    #       that distributes a for loop into equal-sized blocks and allocates them
-    #       to each thread. If this changes, we may need to rethink things here.
-    interval = 1:min(nchains, Threads.nthreads())
-    rngs = [deepcopy(rng) for _ in interval]
     models = [deepcopy(model) for _ in interval]
     samplers = [deepcopy(sampler) for _ in interval]
-
-    # Create a seed for each chain using the provided random number generator.
-    seeds = rand(rng, UInt, nchains)
 
     # Set up a chains vector.
     chains = Vector{Any}(undef, nchains)
@@ -343,10 +349,7 @@ function mcmcsample(
                         # Obtain the ID of the current thread.
                         id = Threads.threadid()
 
-                        # Seed the thread-specific random number generator with the pre-made seed.
                         subrng = rngs[id]
-                        Random.seed!(subrng, seeds[i])
-
                         # Sample a chain and save it to the vector.
                         chains[i] = StatsBase.sample(subrng, models[id], samplers[id], N;
                                                      progress = false, kwargs...)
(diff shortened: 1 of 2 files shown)


src/sample.jl
```diff
@@ -284,8 +284,23 @@ function mcmcsample(
     nchains::Integer;
     progress = PROGRESS[],
     progressname = "Sampling ($(min(nchains, Threads.nthreads())) threads)",
+    rngs::Union{Nothing, AbstractVector{<:Random.AbstractRNG}} = nothing,
```
devmotion, line 287: I think it would be confusing to have both an `rng` argument and a `rngs` keyword argument. Maybe we could just allow a vector of RNGs as the first argument and not use keyword arguments at all?

src/sample.jl
```diff
@@ -284,8 +284,23 @@ function mcmcsample(
...
     kwargs...
 )
+    interval = 1:min(nchains, Threads.nthreads())
```
devmotion, line 290: I think we should not change the order and perform checks e.g. of the number of threads and processes first, before creating the RNGs.

src/sample.jl
```diff
@@ -284,8 +284,23 @@ function mcmcsample(
...
+        # Copy the random number generator, model, and sample for each thread
+        # NOTE: As of May 17, 2020, this relies on Julia's thread scheduling functionality
+        #       that distributes a for loop into equal-sized blocks and allocates them
+        #       to each thread. If this changes, we may need to rethink things here.
+        subrngs = [deepcopy(rng) for _ in interval]
+        # Create a seed for each chain using the provided random number generator.
+        seeds = rand(rng, UInt, nchains)
+    else
+        subrngs = rngs
+        @assert length(rngs) == nchains
```
devmotion, line 301: In principle it would be sufficient to only provide `min(nchains, Threads.nthreads())` RNGs - why do we require `nchains` RNGs?

I think also we should use `error` or e.g. `ArgumentError` with a clear error message instead of `@assert` since we really want to check user input here and not some internal functionality:
```suggestion
        length(rngs) == nchains || error("the number of chains and random number generators has to be equal")
```
Also,

src/sample.jl
```diff
@@ -375,8 +382,18 @@ function mcmcsample(
     nchains::Integer;
     progress = PROGRESS[],
     progressname = "Sampling ($(Distributed.nworkers()) processes)",
+    rngs::Union{Nothing, AbstractVector{<:Random.AbstractRNG}} = nothing,
```
devmotion, line 385: Same as above, I think it is better to allow a vector of RNGs as first argument.

src/sample.jl
```diff
@@ -375,8 +382,18 @@ function mcmcsample(
...
     kwargs...
 )
+    if rngs === nothing
+        rngs = [deepcopy(rng) for _ in 1:nchains]
+        # Create a seed for each chain using the provided random number generator.
+        seeds = rand(rng, UInt, nchains)
+        Random.seed!.(rngs, seeds)
```
devmotion, line 392: There is no need to invoke the broadcasting machinery here.
```suggestion
        foreach(Random.seed!, rngs, seeds)
```

src/sample.jl
```diff
@@ -434,8 +445,10 @@ function mcmcsample(
                         progress && put!(channel, true)
 
                         # Return the new chain.
-                        return chain
+                        return chain, rng
                     end
+                    chains = first.(chains_rngs)
+                    rngs .= last.(chains_rngs)
```
devmotion, line 451: This seems to be an unnecessary change?
```suggestion
                        return chain
                    end
```

test/sample.jl
```diff
@@ -223,6 +223,26 @@
...
+        rngs = [deepcopy(Random.GLOBAL_RNG) for _ in 1:1000]
+        seeds = rand(Random.GLOBAL_RNG, UInt, 1000)
+        Random.seed!.(rngs, seeds)
+ 
+        model, sampler = MyModel(), MySampler()
+        _chains1 = sample(model, sampler, MCMCDistributed(), 5000, 1000; chain_type = MyChain, rngs = rngs)
+        _chains2 = sample(model, sampler, MCMCDistributed(), 5000, 1000; chain_type = MyChain, rngs = rngs)
+        function combine(c1::MyChain, c2::MyChain)
+            MyChain(vcat(c1.as, c2.as), vcat(c1.bs, c2.bs), nothing)
+        end
```
devmotion, line 238: Just use `chainscat` or `chainsstack`?

test/sample.jl
```diff
@@ -223,6 +223,26 @@
                    progress = false, chain_type = MyChain)
         end
         @test all(l.level > Logging.LogLevel(-1) for l in logs)
+
+        # Subrngs
+        Random.seed!(1234)
+        rngs = [deepcopy(Random.GLOBAL_RNG) for _ in 1:1000]
+        seeds = rand(Random.GLOBAL_RNG, UInt, 1000)
+        Random.seed!.(rngs, seeds)
```
devmotion, line 231:
```suggestion
        foreach(Random.seed!, rngs, seeds)
```

test/sample.jl
```diff
@@ -223,6 +223,26 @@
...
+ 
+        model, sampler = MyModel(), MySampler()
+        _chains1 = sample(model, sampler, MCMCDistributed(), 5000, 1000; chain_type = MyChain, rngs = rngs)
```
devmotion, line 234: It seems you only check `MCMCDistributed` but not `MCMCThreads`?

devmotion (changes requested):
My first question, as always, is: Where is this functionality (supposed to be) used and is it common enough to justify a more complicated implementation and API? The default implementations in AbstractMCMC are not supposed to cover all possible use cases but the most common ones (one can implement specialized sample procedures manually with the interface, and e.g. use the iteration or transducer interface).

We should also update the documentation and the version number.

mohamed82008:
>My first question, as always, is: Where is this functionality (supposed to be) used and is it common enough to justify a more complicated implementation and API?

The use case I have in mind is to do chain based sampling pausing and resuming in a reproducible way that's comparable to running the full chain, in serial, multi-threading and multi-processing.

devmotion:
Hmm I'm sorry, I don't understand how you want to make it comparable to pausing and resuming a single chain. Even if you set the RNGs in the multichain methods the samples won't be the same as in the single chain. It also seems you forgot to add this feature in the serial sampling.

Or is your motivation that serial, multithreaded and distributed sampling yield different chains? I'm not sure if this is actually a problem as you can get different chains even for the same method, same RNG, same seed and same Julia version on different architectures (and even more so with different Julia versions). The discrepancy between multithreaded and distributed arises only from the fact that we can use a lower number of RNGs in multithreading and hence reduce the copy overhead. In serial sampling we don't have to sample seeds and hence we deviate from the other methods.

mohamed82008:
Ya I figured a way to do what I want without this PR. Will close this one. Thanks for your feedback.

PR #95, opened 2022-02-19
Remove use of `threadid`

devmotion:
One must not use `threadid` in `@threads for ....`, as discussed in https://discourse.julialang.org/t/behavior-of-threads-threads-for-loop/76042. This PR implements an alternative suggested in https://github.com/JuliaMolSim/DFTK.jl/issues/588.

Edit: Since `@spawn` was introduced in Julia 1.3, we have to drop Julia < 1.3 or add special version guards.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.0'
+          - '1.3'
           - '1'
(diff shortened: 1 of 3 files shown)


cpfiffer (approved):
Love it! Let's gooooooooo

cpfiffer:
Failing tests were known and relate to MCMCChains, I ignored them and merged in.

PR #97, opened 2022-02-21
Ensure ensemble sampling is reproducible

devmotion:
I noticed that in #95 I accidentally introduced a reproducibility issue for multithreaded sampling: If seeds are set for each chunk (as in the PR), then the samples will depend on the number of threads. This PR fixes the issue and reverts this change: For each chain, the RNG is set to a specific reproducible seed.

Additionally, I added the same seeds to the serial sampling (they are already used in the distributed sampling). This ensures that the samples do not depend on the chosen ensemble methods. It could be removed from the PR if not desired.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -46,6 +46,8 @@ jobs:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
       - uses: julia-actions/cache@v1
+        with:
+          cache-packages: "false" # caching Conda.jl causes precompilation error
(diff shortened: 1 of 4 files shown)


cpfiffer (approved):
Ah, good fix and good catch!

cpfiffer:
I rolled the version up because I submitted 3.3.1 too early, woops :disappointed_relieved:

devmotion:
Hmm are you sure? Wasn't 3.3.1 correct? With 3.3.2 automerge complains: https://github.com/JuliaRegistries/General/pull/55232

PR #99, opened 2022-02-22
Fix test errors and simplify docs

devmotion:
This PR applies some useful fixes and simplifications from https://github.com/TuringLang/AbstractMCMC.jl/pull/94.


test/sample.jl
```diff
@@ -310,7 +312,7 @@
     @testset "Sample stats" begin
         chain = sample(MyModel(), MySampler(), 1000; chain_type = MyChain)
         
-        @test chain.stats.stop > chain.stats.start
+        @test chain.stats.stop >= chain.stats.start
```
devmotion, line 315: It seems the resolution of `time()` on Windows + Julia 1.3 can be pretty bad only in the range of ms (don't know if this is Windows and/or Julia 1.3 specific): https://github.com/TuringLang/AbstractMCMC.jl/runs/5294044506?check_suite_focus=true#step:6:587 I.e., even with 1000 samples `stats.stop` and `stats.start` can be equal. Unfortunately, the more idiomatic `@timed` macros caused performance regressions: https://github.com/TuringLang/AbstractMCMC.jl/pull/64#issuecomment-813774289

devmotion:
Bump :slightly_smiling_face:

yebai (approved):
thanks, @devmotion - it looks good!

PR #102, opened 2022-05-29
Fix `discard_initial`, and add support for `discard_initial` and `thinning` to iterator and transducer

devmotion:
This PR fixes a bug in `discard_initial`. Additionally, it adds support for `discard_initial` and `thinning` to the iterator and transducer.

Of course, it would be possible to achieve the same behaviour in the iterator/transducer by using `Iterators.drop`/`Transducers.Drop` etc. but I think support for these two keyword arguments improves the user experience since it is more consistent with `sample` and arguably simpler.

The PR also adds tests for these two keyword arguments.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "4.1.0"
+version = "4.1.1"
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -53,7 +53,7 @@ Common keyword arguments for regular and parallel sampling are:
 - `thinning` (default: `1`): factor by which to thin samples.
 
 !!! info
-    The common keyword arguments `progress`, `chain_type`, and `callback` are not supported by the iterator [`AbstractMCMC.steps`](@ref) and the transducer [`AbstractMCMC.Samples`](@ref).
+    The common keyword arguments `progress`, `chain_type`, and `callback` are not supported by the iterator [`AbstractMCMC.steps`](@ref) and the transducer [`AbstractMCMC.Sample`](@ref).
 
 There is no "official" way for providing initial parameter values yet.
 However, multiple packages such as [EllipticalSliceSampling.jl](https://github.com/TuringLang/EllipticalSliceSampling.jl) and [AdvancedMH.jl](https://github.com/TuringLang/AdvancedMH.jl) support an `init_params` keyword argument for setting the initial values when sampling a single chain.
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -43,8 +43,7 @@ AbstractMCMC.MCMCSerial
 
 ## Common keyword arguments
 
-Common keyword arguments for regular and parallel sampling (not supported by the iterator and transducer)
-are:
+Common keyword arguments for regular and parallel sampling are:
 - `progress` (default: `AbstractMCMC.PROGRESS[]` which is `true` initially):  toggles progress logging
 - `chain_type` (default: `Any`): determines the type of the returned chain
 - `callback` (default: `nothing`): if `callback !== nothing`, then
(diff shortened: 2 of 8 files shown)


cpfiffer (approved):

devmotion:
The test errors are quite puzzling. It seems changing the reference sampling to `progress=true` fixes them locally with Julia 1.3. I don't know why enabling progress logs would affect the samples on Julia 1.3.

cpfiffer:
Yeah the reason for failure here is really unclear. Can you print out the full chains for the two chains with something like 5 samples or something? Maybe it'll be obvious to us if there's some kind of index arithmetic error.

devmotion:
I think it could be related to some changes in Julia (logging? the thread-local rngs?) since tests fail on Julia 1.3 but not on more recent versions.

The PR adds some tests that checks that the bug with `discard_initial` is fixed. E.g., with the master branch currently one gets the following on Julia 1.7 since not `discard_initial` but only `max(0, discard_initial - 1)` samples are discarded:
```julia
julia> using AbstractMCMC

julia> using Random: Random, AbstractRNG

julia> using Test

julia> include(joinpath(dirname(dirname(pathof(AbstractMCMC))), "test", "utils.jl"))

julia> Random.seed!(1234);

julia> N = 100;

julia> discard_initial = 50;

julia> chain = sample(MyModel(), MySampler(), N; discard_initial=discard_initial);

julia> @test length(chain) == N
Test Passed
  Expression: length(chain) == N
   Evaluated: 100 == 100

julia> @test !ismissing(chain[1].a)
Test Passed
  Expression: !(ismissing((chain[1]).a))

julia> # Repeat sampling without discarding initial samples.
       Random.seed!(1234);

julia> ref_chain = sample(MyModel(), MySampler(), N + discard_initial; progress=false);

julia> [chain[i].a for i in 1:5]
5-element Vector{Float64}:
 0.9327827434839276
 0.4122251300478674
 0.7553435356857371
 0.7091580778375822
 0.18668155849545398

julia> [ref_chain[i + discard_initial].a for i in 1:5]
5-element Vector{Float64}:
 0.4122251300478674
 0.7553435356857371
 0.7091580778375822
 0.18668155849545398
 0.6932923170086805
```
Whereas with this PR one gets on Julia 1.7:
```julia
...

julia> [chain[i].a for i in 1:5]
5-element Vector{Float64}:
 0.4122251300478674
 0.7553435356857371
 0.7091580778375822
 0.18668155849545398
 0.6932923170086805

julia> [ref_chain[i + discard_initial].a for i in 1:5]
5-element Vector{Float64}:
 0.4122251300478674
 0.7553435356857371
 0.7091580778375822
 0.18668155849545398
 0.6932923170086805
```

However, on Julia 1.3 it seems completely messed up (this PR):
```julia
...

julia> [chain[i].a for i in 1:5]
5-element Array{Float64,1}:
 0.9906023534355743 
 0.10101494660193899
 0.43508662726114   
 0.8837184408191245 
 0.07624588897736362

julia> [ref_chain[i + discard_initial].a for i in 1:5]
5-element Array{Float64,1}:
 0.26559536159694797
 0.6126276964747348 
 0.5083634348868193 
 0.6125537989226799 
 0.8510698470611828 
```

However, enabling the progress logging for the reference trajectory (i.e., using the same setting for the chain and the reference) fixes the problem:
```julia
...

julia> Random.seed!(1234);

julia> ref_chain = sample(MyModel(), MySampler(), N + discard_initial; progress=true);

julia> [ref_chain[i + discard_initial].a for i in 1:5]
5-element Array{Float64,1}:
 0.9906023534355743 
 0.10101494660193899
 0.43508662726114   
 0.8837184408191245 
 0.07624588897736362
```

It's still mysterious to me why and how progress logging interferes with the RNG, and why this happens only on older Julia versions.

devmotion:
OK, even more surprising - the problem can be fixed on Julia 1.3 by specifying an RNG explicitly, even if the reference trajectory is sampled with `progress=false`:
```julia
...

julia> rng = Random.MersenneTwister(1234);

julia> chain = sample(rng, MyModel(), MySampler(), N; discard_initial=discard_initial);

julia> rng = Random.MersenneTwister(1234);

julia> ref_chain = sample(rng, MyModel(), MySampler(), N + discard_initial; progress=false);

julia> [chain[i].a for i in 1:5]
5-element Array{Float64,1}:
 0.26559536159694797
 0.6126276964747348 
 0.5083634348868193 
 0.6125537989226799 
 0.8510698470611828 

julia> [ref_chain[i + discard_initial].a for i in 1:5]
5-element Array{Float64,1}:
 0.26559536159694797
 0.6126276964747348 
 0.5083634348868193 
 0.6125537989226799 
 0.8510698470611828 
```

cpfiffer:
This sounds like we should just set `progress=true`. I'm not particularly inclined to modify code to suit pre-1.6, given that 1.6 is the current LTS. Alternatively we could even just drop 1.3 testing entirely?

devmotion:
> This sounds like we should just set `progress=true`.

OK, I did that.

> I'm not particularly inclined to modify code to suit pre-1.6, given that 1.6 is the current LTS.

I'm not so much worried about < 1.6 but rather that the issue might reappear due to some internal changes in UUIDs or ProgressLogging. But I guess changing the tests is fine for now and we can reconsider if there's an issue with more recent Julia versions as well.

> Alternatively we could even just drop 1.3 testing entirely?

In my experience it's easy to introduce unsupported syntax or functions and thereby break older Julia versions if they are not tested. Hence I tend to test the oldest officially declared to be compatible version, LTS, and the latest stable release.

devmotion:
I don't understand the test error with Julia nightly on Linux 32bit - it works in all other Julia versions and is caused by a functionality (serial sampling) that is not affected by this PR. I'll ignore it for now.

cpfiffer:
Thank you!

PR #103, opened 2022-06-01
More stable tests

devmotion:
It seems there are still random test errors that are difficult to reproduce. It seems replacing `===` with `==` was helpful, so this PR changes it in more places (`===` was used since the initial sample of `a` is `missing` and can't be compared with `==`).


cpfiffer (approved):
Love it. Could you increment the patch number too?

PR #104, opened 2022-06-01
Replace `GLOBAL_RNG` with `default_rng()`

devmotion:
In Julia >= 1.3, actually `Random.default_rng()` is the intended (and now also properly documented: https://github.com/JuliaLang/julia/pull/44733) way to obtain the default RNG.


cpfiffer (approved):
What is CI good for, anyway

PR #106, opened 2022-08-26
Update CompatHelper.yml

devmotion:
CompatHelper started to fail since it pulls in an old version. I updated the action with the implementation suggested by CompatHelper.jl.


devmotion:
I triggered the action manually on this branch to see if it actually works: https://github.com/TuringLang/AbstractMCMC.jl/actions/runs/2932210779

devmotion:
Seems to work.

PR #110, opened 2022-12-08
Add AbstractModel implementation of LogDensityProblems interface

torfjelde:
This PR adds an implementation of `AbstractModel` which supports the LogDensityProblems.jl interface.

Benefits:
1. We get a good "default" implementation for a `AbstractModel`, allowing sharing of model definitions between different sampler packages, e.g. with this AdvancedHMC.jl (see https://github.com/TuringLang/AdvancedHMC.jl/pull/301 for related discussion) and AdvancedMH.jl can both use `LogDensityModel` instead of defining their own.
2. We get AD "for free" through LogDensityProblemsAD.jl.


Project.toml
```diff
@@ -3,12 +3,13 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "4.1.3"
+version = "4.1.4"
```
devmotion, line 6: This should be
```suggestion
version = "4.2"
```
shouldn't it? It's a new feature, not a bugfix.

src/logdensityproblems.jl
```diff
@@ -0,0 +1,23 @@
...
+end
+function LogDensityProblems.capabilities(model::LogDensityModel)
+    return LogDensityProblems.capabilities(model.logdensity)
+end
+function LogDensityProblems.logdensity(model::LogDensityModel, x)
+    return LogDensityProblems.logdensity(model.logdensity, x)
+end
+function LogDensityProblems.logdensity_and_gradient(model::LogDensityModel, x)
+    return LogDensityProblems.logdensity_and_gradient(model.logdensity, x)
+end
```
devmotion, line 23: My only concern is that wrappers can be very annoying due to method ambiguity issues (if e.g. someone defines methods for specific x and some abstract logdensity type, but is not aware of `LogDensityModel`) and since you are screwed if you forget to add some definition that should be forwarded (or some log densities define additional functions).

Maybe don't define any `LogDensityProblems` functions at all? If you know it's a LogDensityModel, you can just call `dimension(model.logdensity)` etc. in your sampler. And if not, you can't call `dimension` etc. on your model anyway.

So to summarize: I don't think `LogDensityProblems.dimension` etc. should be defined for the model.

Project.toml
torfjelde: Oh true :+1:

src/logdensityproblems.jl
torfjelde: There is no abstract logdensity type, right? So If someone implements something like `LogDensityProblems.dimension`, it should always be a type they've defined themselves.

devmotion: My point is different (I think): I think it is sufficient if LogDensityProblems.dimension etc. are defined for `model.logdensity` and don't see the need for defining them for `model`. That would also avoid the issues with the wrapper.

torfjelde: I mean, yeah, you definitively can do without it, but IMO it can be annoying.

Most of the time you don't use AbstractMCMC in isolation, and so IMO it's nice to go

``` julia
model = AbstractMCMC.LogDensityModel(logdensitymodel)

smaples = sample(model, ...)

something_else_which_requires_logdensityproblems_interface(model, ...)
```

instead of

``` julia
model = AbstractMCMC.LogDensityModel(logdensitymodel)

smaples = sample(model, ...)

# User has to manually unwrap to call the method.
something_else_which_requires_logdensityproblems_interface(model.logdensity, ...)
```

It's definitively not a big thing, _but_ I can't see the _downside_ of implementing these methods (since method ambiguities won't be an issue, as described above, unless someone is doing type-piracy).

devmotion: It will be an issue as soon as someone defines e.g. `logdensity(model::AbstractModel, x::MyType)` I think? And the new LogDensityProblemsAD is another issue: You would have to depend on it here as well if you want to support `ADgradient(model::LogDensityModel)` (which IMO for the same reasons we should not do).

In any case, I don't see a good reason why you *should* add these methods: If you work with `LogDensityModel` in a sampler, you already have to restrict your dispatches to this specific type of models because otherwise you can't call `logdensity` etc. So you know you only accept this type anyway, and then you can just as easily call `dimension(model.logdensity)` in your sampler.

The example I think is also not really a problem: In that case the user already has the `model.logdensity` since the `model` is constructed with it in the first place :slightly_smiling_face: And even if the user does not, I don't see a problem with using `model.logdensity` instead of `model` since it allows us to avoid all these wrapper-related issues.

torfjelde: But would `logdensity(model::AbstractModel, x::MyType)` ever happen, if we say "if you're working with `LogDensityProblems`, use `LogDenstityModel`"? 

> In that case the user already has the model.logdensity since the model is constructed with it in the first place slightly_smiling_face And even if the user does not, I don't see a problem with using model.logdensity instead of model since it allows us to avoid all these wrapper-related issues.

It's not a _problem_, it's just _annoying_ :upside_down_face:  In particular for users that aren't too familiar with Julia. I really like the idea of being able to tell these users "Hey, if you don't want to use Turing but you want to use the samplers for your own model, just implement LogDensityInterface and wrap your model in `LogDensityModel`; all your existing code that assumes LogDensityProblems.jl-interface will work nicely with this too, so you lose nothing."

_But_ I'm not married to the idea! I am okay going your route too :+1:

cpfiffer: I think my sense of this is that I can't imagine that case where someone is calling `logdensity(model::AbstractModel, x::MyType)` means they do not want to use any of the `LogDensityModel` stuff, and that's fine -- they might be a little confused at first if they are not familiar, but this seems restrictive enough that it's not going to cause much difficulty. 

I am also in favor of the simplicity of the wrappers since dot notation hurts me, and I'd rather use a functional extractor.

torfjelde: What say you @devmotion ?

devmotion: > I really like the idea of being able to tell these users "Hey, if you don't want to use Turing but you want to use the samplers for your own model, just implement LogDensityInterface and wrap your model in LogDensityModel; all your existing code that assumes LogDensityProblems.jl-interface will work nicely with this too, so you lose nothing."

Wouldn't it be even cooler and cleaner if it would just work automatically and you don't have to wrap your LogDensityProblem-compatible model? Maybe we should support arbitrary types of models that implement the LogDensityProblems interface in AbstractMCMC and downstream packages? If that's the general API then one could just sample with anything and always call `logdensity` etc, regardless of the type (if samplers want to support that - otherwise they could still restrict `sample` to some specific types of models).

A `LogDensityModel` would be mainly useful then if multiple sampler packages have the same structure of models. But its purpose would not be to forward `logdensity` to `logdensity` of some wrapped function/model - since in that case you could have used the wrapped function/model directly.

torfjelde: > Wouldn't it be even cooler and cleaner if it would just work automatically and you don't have to wrap your LogDensityProblem-compatible model?

Aye it would be, but wouldn't `sample(model, ...)` without restricting to `AbstractModel` be doomed to lead to method ambiguities? 

> A LogDensityModel would be mainly useful then if multiple sampler packages have the same structure of models. But its purpose would not be to forward logdensity to logdensity of some wrapped function/model - since in that case you could have used the wrapped function/model directly.

As I said, I'm happy to drop them and just get this `LogDensityModel` into the package without the forwarded methods, with its purpose just being to indicate that the wrapped object implements the `LogDensityInterface`. So should I just do this then?

devmotion: > Aye it would be, but wouldn't sample(model, ...) without restricting to AbstractModel be doomed to lead to method ambiguities?

Is there a specific example you have in mind? I think it should be fine eg. if we define no method with type restriction on model in AbstractMCMC (as other arguments should also be more specific in downstream implementations) but maybe there's an example that would be problematic. Generally, I assume due to the AbstractSampler argument and the fact that StatsBase mainly deals with sampling arrays (IIRC) there should not be any problems with other (common) packages - but again, maybe I miss something.

> As I said, I'm happy to drop them and just get this LogDensityModel into the package without the forwarded methods, with its purpose just being to indicate that the wrapped object implements the LogDensityInterface. So should I just do this then?

I think this would be the best way forward - not only because I still have doubts regarding the interface for LogDensityProblems but also because IMO it's easier to add stuff incrementally than to remove and break functionality and interfaces again.

torfjelde: > Is there a specific example you have in mind?

No, just slightly worried it might happen, though I agree I'd expect it to generally be okay.

> I think this would be the best way forward - not only because I still have doubts regarding the interface for LogDensityProblems but also because IMO it's easier to add stuff incrementally than to remove and break functionality and interfaces again.

I agree with this :+1: I'll do that then:)

torfjelde: Made the change; could you have a look now?:)

src/logdensityproblems.jl
```diff
@@ -0,0 +1,15 @@
...
+Wrapper around something that implements the LogDensityProblem.jl interface.
+
+Note that this does _not_ implement the LogDensityProblems.jl interface itself,
+but it simply useful for indicating to the `sample` and other `AbstractMCMC` methods
+that the wrapped object implements the LogDensityProblems.jl interface.
+
+# Fields
+- `logdensity`: The object that implements the LogDensityProblems.jl interface.
+"""
+struct LogDensityModel{L} <: AbstractModel
```
yebai, line 13: Maybe make `L` a subtype of `LogDensityProblems`? Otherwise, we import but never use `LogDensityProblems.jl`.

Project.toml
```diff
@@ -3,12 +3,13 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
...
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "4.1.3"
+version = "4.2"
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
 ConsoleProgressMonitor = "88cd18e8-d9cc-4ea6-8889-5259c0d15c8b"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
+LogDensityProblems = "6fdf6af0-433a-55f7-b3ed-c6c6e0b8df7c"
```
devmotion, line 12: That's not required if we don't implement the LogDensityProblems interface, it seems.
```suggestion
```

Project.toml
```diff
@@ -20,6 +21,7 @@ Transducers = "28d57a85-8fef-5791-bfe6-a80928e7c999"
 [compat]
 BangBang = "0.3.19"
 ConsoleProgressMonitor = "0.1"
+LogDensityProblems = "2"
```
devmotion, line 24:
```suggestion
```

src/AbstractMCMC.jl
```diff
@@ -7,6 +7,7 @@ using ProgressLogging: ProgressLogging
 using StatsBase: StatsBase
 using TerminalLoggers: TerminalLoggers
 using Transducers: Transducers
+using LogDensityProblems: LogDensityProblems
```
devmotion, line 10:
```suggestion
```

src/logdensityproblems.jl
torfjelde: There's no `LogDensityProblem` abstract type, so I'll just do what @devmotion suggested:)

torfjelde:
Do we still support Julia 1.3?

torfjelde:
And we should at least be testing on Julia 1.6, which we currently are not doing.

torfjelde:
As long as you're happy with the bump to 1.6, this should now be ready to go:)

yebai (approved):

devmotion (approved):

PR #113, opened 2022-12-30
Support log density functions as models

devmotion:
This PR adds implementations of `sample` for generic log density functions. It is assumed that models in the `sample` calls that are not subtypes of `AbstractModel` are log density functions. They are supposed to support the LogDensityProblems interface (this is checked) and wrapped in a `LogDensityModel` automatically, which is then used instead of the log density in the `sample` call.

The motivation is that in downstream PRs to me it seemed a bit inconvenient that users have to wrap their log density functions manually. By adding aupport for automatic wrapping to AbatractMCMC instead of downstream packages, all downstream packages can benefit from it without additional code in downstream packages (they only have to support LogDensityModel).

A potential problem could be method ambiguity issues. However, since AbstractSampler is owned by AbstractMCMC this would only be partly AbstractMCMC's fault and I assume it won't happen in practice (let's wait for the downstream tests).


src/sample.jl
```diff
@@ -85,6 +120,36 @@ function StatsBase.sample(
     return mcmcsample(rng, model, sampler, parallel, N, nchains; kwargs...)
 end
 
+# Fallback: Wrap log density function in a model
+"""
+    sample(
+        rng::Random.AbstractRNG=Random.default_rng(),,
```
torfjelde, line 126:
```suggestion
        rng::Random.AbstractRNG=Random.default_rng(),
```

torfjelde (approved):
LGTM if downstream tests are passing! I only have one change relating to a typo in the docstring, but otherwise es bueno.

devmotion:
I wonder if we could/should add some tests in AbstractMCMC, in addition to the existing downstream tests.

yebai (approved):

torfjelde:
Though I agree tests would be ideal, if you're busy atm I'm happy to
merge as is given that downstream tests pass.

cpfiffer (approved):
Looks like a net positive to me, and for relatively low cost. I say we merge it in!

torfjelde:
Oh actually one question @devmotion : do we need `LogDensityModel` after this PR?

devmotion:
I would say, yes, it's still useful since it allows downstream packages to clearly state the `AbstractModel`-type(s) that they want to support with their sampler(s). Additionally, a dedicated model type for log densities makes it easier to define conversion from/to other model types.

devmotion:
I'll see if I find time to add some tests tomorrow. Otherwise I'll merge on Monday, if there are no objections.

devmotion:
I added tests but let's merge https://github.com/TuringLang/AbstractMCMC.jl/pull/114 first (I noticed CompatHelper was disabled due to inactivity).

torfjelde (approved):
LGTM! Happy to merge as soon as tests pass. Awesome stuff @devmotion :)

PR #115, opened 2023-01-10
Do not test Atom/Juno

devmotion:
This PR removes tests of progress logging with Atom/Juno since it holds back other test dependencies: The package is in maintenance-only mode and CompatHelper PRs are not merged anymore (see https://github.com/JunoLab/Atom.jl/pulls).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -30,10 +30,9 @@ Transducers = "0.4.30"
 julia = "1.6"
 
 [extras]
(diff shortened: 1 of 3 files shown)


cpfiffer (approved):
Sad, but a good idea since Juno is sunsetting.

cpfiffer:
I know! Just remarking that removing tests is probably a good idea anyway, following the general trend of Juno.

PR #117, opened 2023-03-09
Addition of `step_warmup`

torfjelde:
For many samplers, it might be useful to separate between the warmup phase and the sampling phase, e.g. in AdvancedHMC we have an initial phase where we adapt the parameters to the parameters at hand.

Currently, the usual approach to implementing such a warmup stage is to keep track of the iteration + the adaptation stuff internally in the state, but sometimes that can be quite annoying and/or redundant to implement.

I would also argue that separating these is useful on a conceptual level, e.g. even if I interact directly with the stepper-interface, I would now do

``` julia
for _ = 1:nwarmup
    state = last(AbstractMCMC.step_warmup(rng, model, sampler, state))
end

for i = 1:nsteps
    transition, state = AbstractMCMC.step(rng, model, sampler, state)
    # save
    ...
end
```

vs.

``` julia
for i = 1:nwarmup + nsteps
    transition, state = AbstractMCMC.step(rng, model, sampler, state)
    # save
    if !iswarmup(state)
        ...
    end
end
```

With this PR, for something like MCMCTempering.jl where in the warmup phase we actually just want to take steps with the underlying sampler rather than also include the swaps, we can then just make the `step_warmup` do so without having to add any notion of iterations in the state, nor without telling the sampler itself about how many warm-up steps we want (instead it's just specified by kwarg in `sample`, as it should be).

### DIFF
--- a/docs/src/design.md
+++ b/docs/src/design.md
@@ -69,7 +69,7 @@ If one also has some special handling of the warmup-stage of sampling, then this
 AbstractMCMC.step_warmup
 ```
 
-which will be used for the first `num_warmup`, as specified as a keyword argument to [`AbstractMCMC.sample`](@ref). 
+which will be used for the first `num_warmup` iterations, as specified as a keyword argument to [`AbstractMCMC.sample`](@ref). 
 Note that this is optional; by default it simply calls [`AbstractMCMC.step`](@ref) from above.
 
 ## Collecting samples
--- a/docs/src/design.md
+++ b/docs/src/design.md
@@ -63,6 +63,14 @@ the sampling step of the inference method.
 AbstractMCMC.step
 ```
 
+If one also has some special handling of the warmup-stage of sampling, then this can be specified by overloading
+
+```@docs
+AbstractMCMC.step_warmup
+```
+
+Note that this is optional; by default it simply calls [`AbstractMCMC.step`](@ref) from above.
+
 ## Collecting samples
 
 !!! note
--- a/docs/src/design.md
+++ b/docs/src/design.md
@@ -69,6 +69,7 @@ If one also has some special handling of the warmup-stage of sampling, then this
 AbstractMCMC.step_warmup
 ```
 
+which will be used for the first `num_warmup`, as specified as a keyword argument to [`AbstractMCMC.sample`](@ref). 
 Note that this is optional; by default it simply calls [`AbstractMCMC.step`](@ref) from above.
 
 ## Collecting samples
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -57,6 +57,19 @@ current `state` of the sampler.
 function step end
 
 """
+    step_warmup(rng, model, sampler[, state; kwargs...])
+
+Return a 2-tuple of the next sample and the next state of the MCMC `sampler` for `model`.
+
+When sampling using [`sample`](@ref), this takes the place of [`step`](@ref) in the first `discard_initial`
+number of iterations. This is useful if the sampler has a "warmup"-stage initial stage
+that is different from the standard iteration.
+
+By default, this simply calls [`step`](@ref.)
+"""
+step_warmup(rng, model, sampler, state; kwargs...) = step(rng, model, sampler, state; kwargs...)
+
+"""
     samples(sample, model, sampler[, N; kwargs...])
 
 Generate a container for the samples of the MCMC `sampler` for the `model`, whose first
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -61,9 +61,10 @@ function step end
 
 Return a 2-tuple of the next sample and the next state of the MCMC `sampler` for `model`.
 
-When sampling using [`sample`](@ref), this takes the place of [`step`](@ref) in the first `discard_initial`
-number of iterations. This is useful if the sampler has a "warmup"-stage initial stage
-that is different from the standard iteration.
+When sampling using [`sample`](@ref), this takes the place of [`step`](@ref) in the first
+`num_warmup` number of iterations, as specified by the `num_warmup` keyword to [`sample`](@ref).
+This is useful if the sampler has a "warmup"-stage initial stage that is different from the
+standard iteration.
 
 By default, this simply calls [`step`](@ref.)
 """
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -68,7 +68,9 @@ standard iteration.
 
 By default, this simply calls [`step`](@ref.)
 """
-step_warmup(rng, model, sampler, state; kwargs...) = step(rng, model, sampler, state; kwargs...)
+function step_warmup(rng, model, sampler, state; kwargs...)
+    return step(rng, model, sampler, state; kwargs...)
+end
 
 """
     samples(sample, model, sampler[, N; kwargs...])
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -30,13 +30,30 @@ be specified with the `chain_type` argument.
 By default, this method returns `samples`.
 """
 function bundle_samples(
-    samples, ::AbstractModel, ::AbstractSampler, ::Any, ::Type; kwargs...
+    samples, model::AbstractModel, sampler::AbstractSampler, state, ::Type{T}; kwargs...
+) where {T}
+    # dispatch to internal method for default implementations to fix
+    # method ambiguity issues (see #120)
+    return _bundle_samples(samples, model, sampler, state, T; kwargs...)
+end
+
+function _bundle_samples(
+    samples,
+    @nospecialize(::AbstractModel),
+    @nospecialize(::AbstractSampler),
+    @nospecialize(::Any),
+    ::Type;
+    kwargs...,
 )
     return samples
 end
-
-function bundle_samples(
-    samples::Vector, ::AbstractModel, ::AbstractSampler, ::Any, ::Type{Vector{T}}; kwargs...
+function _bundle_samples(
+    samples::Vector,
+    @nospecialize(::AbstractModel),
+    @nospecialize(::AbstractSampler),
+    @nospecialize(::Any),
+    ::Type{Vector{T}};
+    kwargs...,
 ) where {T}
     return map(samples) do sample
         convert(T, sample)
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -61,12 +61,12 @@ function step end
 
 Return a 2-tuple of the next sample and the next state of the MCMC `sampler` for `model`.
 
-When sampling using [`sample`](@ref), this takes the place of [`step`](@ref) in the first
+When sampling using [`sample`](@ref), this takes the place of [`AbstractMCMC.step`](@ref) in the first
 `num_warmup` number of iterations, as specified by the `num_warmup` keyword to [`sample`](@ref).
-This is useful if the sampler has a "warmup"-stage initial stage that is different from the
+This is useful if the sampler has an initial "warmup"-stage that is different from the
 standard iteration.
 
-By default, this simply calls [`step`](@ref.)
+By default, this simply calls [`AbstractMCMC.step`](@ref).
 """
 step_warmup(rng, model, sampler; kwargs...) = step(rng, model, sampler; kwargs...)
 step_warmup(rng, model, sampler, state; kwargs...) = step(rng, model, sampler, state; kwargs...)
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -86,7 +86,9 @@ standard iteration.
 By default, this simply calls [`AbstractMCMC.step`](@ref).
 """
 step_warmup(rng, model, sampler; kwargs...) = step(rng, model, sampler; kwargs...)
-step_warmup(rng, model, sampler, state; kwargs...) = step(rng, model, sampler, state; kwargs...)
+function step_warmup(rng, model, sampler, state; kwargs...)
+    return step(rng, model, sampler, state; kwargs...)
+end
 
 """
     samples(sample, model, sampler[, N; kwargs...])
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -68,9 +68,8 @@ standard iteration.
 
 By default, this simply calls [`step`](@ref.)
 """
-function step_warmup(rng, model, sampler, state; kwargs...)
-    return step(rng, model, sampler, state; kwargs...)
-end
+step_warmup(rng, model, sampler; kwargs...) = step(rng, model, sampler; kwargs...)
+step_warmup(rng, model, sampler, state; kwargs...) = step(rng, model, sampler, state; kwargs...)
 
 """
     samples(sample, model, sampler[, N; kwargs...])
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -12,32 +12,29 @@ function setprogress!(progress::Bool)
     return progress
 end
 
-function StatsBase.sample(model::AbstractModel, sampler::AbstractSampler, arg; kwargs...)
-    return StatsBase.sample(Random.default_rng(), model, sampler, arg; kwargs...)
-end
-
-"""
-    sample([rng, ]model, sampler, N; kwargs...)
-
-Return `N` samples from the `model` with the Markov chain Monte Carlo `sampler`.
-"""
 function StatsBase.sample(
-    rng::Random.AbstractRNG,
-    model::AbstractModel,
-    sampler::AbstractSampler,
-    N::Integer;
-    kwargs...,
+    model_or_logdensity, sampler::AbstractSampler, N_or_isdone; kwargs...
 )
-    return mcmcsample(rng, model, sampler, N; kwargs...)
+    return StatsBase.sample(
+        Random.default_rng(), model_or_logdensity, sampler, N_or_isdone; kwargs...
+    )
 end
 
 """
-    sample([rng, ]model, sampler, isdone; kwargs...)
+    sample(
+        rng::Random.AbatractRNG=Random.default_rng(),
+        model::AbstractModel,
+        sampler::AbstractSampler,
+        N_or_isdone;
+        kwargs...,
+    )
 
-Sample from the `model` with the Markov chain Monte Carlo `sampler` until a
-convergence criterion `isdone` returns `true`, and return the samples.
+Sample from the `model` with the Markov chain Monte Carlo `sampler` and return the samples.
 
-The function `isdone` has the signature
+If `N_or_isdone` is an `Integer`, exactly `N_or_isdone` samples are returned.
+
+Otherwise, sampling is performed until a convergence criterion `N_or_isdone` returns `true`.
+The convergence criterion has to be a function with the signature
 ```julia
 isdone(rng, model, sampler, samples, state, iteration; kwargs...)
 ```
@@ -48,14 +45,14 @@ function StatsBase.sample(
     rng::Random.AbstractRNG,
     model::AbstractModel,
     sampler::AbstractSampler,
-    isdone;
+    N_or_isdone;
     kwargs...,
 )
-    return mcmcsample(rng, model, sampler, isdone; kwargs...)
+    return mcmcsample(rng, model, sampler, N_or_isdone; kwargs...)
 end
 
 function StatsBase.sample(
-    model::AbstractModel,
+    model_or_logdensity,
     sampler::AbstractSampler,
     parallel::AbstractMCMCEnsemble,
     N::Integer,
@@ -63,12 +60,20 @@ function StatsBase.sample(
     kwargs...,
 )
     return StatsBase.sample(
-        Random.default_rng(), model, sampler, parallel, N, nchains; kwargs...
+        Random.default_rng(), model_or_logdensity, sampler, parallel, N, nchains; kwargs...
     )
 end
 
 """
-    sample([rng, ]model, sampler, parallel, N, nchains; kwargs...)
+    sample(
+        rng::Random.AbstractRNG=Random.default_rng(),
+        model::AbstractModel,
+        sampler::AbstractSampler,
+        parallel::AbstractMCMCEnsemble,
+        N::Integer,
+        nchains::Integer;
+        kwargs...,
+    )
 
 Sample `nchains` Monte Carlo Markov chains from the `model` with the `sampler` in parallel
 using the `parallel` algorithm, and combine them into a single chain.
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -288,7 +288,6 @@ function mcmcsample(
         # Save the sample.
         samples = AbstractMCMC.samples(sample, model, sampler; kwargs...)
         samples = save!!(samples, sample, 1, model, sampler; kwargs...)
-        
         i = 2
         while !isdone(rng, model, sampler, samples, state, i; progress=progress, kwargs...)
             # Discard thinned samples.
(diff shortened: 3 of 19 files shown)


devmotion (commented):
I think, as the HMC code in Turing, this conflates warmup stages for the sampler with discarding initial samples. In the first case, (usually) you also want to discard these samples but you might even want to discard more samples even after tuning hyperparameters of a sampler.

I also wonder a bit whether AbstractMCMC is the right level for such an abstraction. Or whether, eg it could be done in AdvancedHMC.

torfjelde:
> I think, as the HMC code in Turing, this conflates warmup stages for the sampler with discarding initial samples. In the first case, (usually) you also want to discard these samples but you might even want to discard more samples even after tuning hyperparameters of a sampler.

Oh I definitively agree with this. IMO this wouldn't be a catch-all solution, and you could still argue that in the case of HMC we should stick to the current approach.

But in _many_ cases this sort of behavior would indeed be what the user expects (though I agree with you, I also don't want to remove the allowance of the current `discard_initial` behavior). 

Is it worth introducing a new keyword argument then? Something that is separate from `discard_initial`, allowing you to define a "burn-in" period, and then a separate `num_warmup` that has this potentially special behavior?

> I also wonder a bit whether AbstractMCMC is the right level for such an abstraction. Or whether, eg it could be done in AdvancedHMC.

I don't think we should deal with the adaptation, etc. itself in AbstractMCMC, but there are sooo many samplers that have some form of initial adaptation that it's IMO worth providing a simple hook that let's people do this.

devmotion:
> Is it worth introducing a new keyword argument then? Something that is separate from discard_initial, allowing you to define a "burn-in" period, and then a separate num_warmup that has this potentially special behavior?

Yes, I think we should keep these options separate. I wonder if `discard_initial` should apply to both these warmup stages and the potential burn-in period to be able to keep warm-up samples as well, if desired. Or are we absolutely certain that you would never want to inspect these samples?

devmotion:
Yes, I think these would be reasonable default values.

torfjelde:
Doneso :+1:

torfjelde:
Aight, I've made an attempt at allowing the desired interaction between `discard_initial` and `num_warmup`, but it _does_ complicate the `mcmcsample` a fair bit :confused:

I've also added some docstring for `mcmcsample`. IMO this should be in the documentation as it specifies the default kwargs that will work with all implementers of `step`. Currently there's no way to figure out that `discard_initial` is a thing.

devmotion:
> IMO this should be in the documentation

The standard keyword arguments are listed and explained in the documentation: https://turinglang.org/AbstractMCMC.jl/dev/api/#Common-keyword-arguments

src/sample.jl
```diff
@@ -91,7 +91,29 @@ function StatsBase.sample(
...
+- `progressname`: the name of the progress bar. Defaults to `"Sampling"`.
+- `callback`: a function that is called after each [`AbstractMCMC.step`](@ref).
+   Defaults to `nothing`.
+- `num_warmup`: number of warmup samples to draw. Defaults to `0`.
+- `discard_initial`: number of initial samples to discard. Defaults to `num_warmup`.
+- `thinning`: number of samples to discard between samples. Defaults to `1`.
+- `chain_type`: the type to pass to [`AbstractMCMC.bundle_samples`](@ref) at the
+  end of sampling to wrap up the resulting samples nicely. Defaults to `Any`.
+- `kwargs...`: Additional keyword arguments to pass on to [`AbstractMCMC.step`](@ref).
+"""
```
devmotion, line 116: I think nobody will look up the docstring for the unexported `mcmcsample` function, so it feels listing and explaining keyword arguments in https://turinglang.org/AbstractMCMC.jl/dev/api/#Common-keyword-arguments is the better approach? And possibly extending the docstring of `sample`?

src/sample.jl
```diff
@@ -100,14 +122,21 @@ function mcmcsample(
...
+    num_warmup=0,
+    discard_initial=num_warmup,
     thinning=1,
     chain_type::Type=Any,
     kwargs...,
 )
     # Check the number of requested samples.
     N > 0 || error("the number of samples must be ≥ 1")
-    Ntotal = thinning * (N - 1) + discard_initial + 1
+    Ntotal = thinning * (N - 1) + discard_initial + num_warmup + 1
```
devmotion, line 133: Is this correct? Shouldn't it just stay the same, possibly with some additional checks:
```suggestion
    discard_initial >= 0 || throw(ArgumentError("number of discarded samples must be non-negative"))
    num_warmup >= 0 || throw(ArgumentError("number of warm-up samples must be non-negative"))
    Ntotal = thinning * (N - 1) + discard_initial + 1
    Ntotal >= num_warmup || throw(ArgumentError("number of warm-up samples exceeds the total number of samples"))
```
I thought we would do the following:
- If `num_warmup = 0`, we just do the same as currently: Sample `discard_initial` samples that are discarded + the `N` samples that are returned, possibly after thinning them.
- If `num_warmup > 0`, we still return `N` samples but depending on `discard_initial` part of the `N` samples might be samples from the warm-up stage. For instance:
  - If `num_warmup = 10`, `discard_initiial = 0`, and `N = 100`, we would sample in total `N` samples and return them, whereof the first 10 are warm-up samples.
  - If `num_warmup = 10`, `discard_initial = 10` (the default if you just specify `num_warmup`), and `N = 100`, then we would sample `N + discard_initial = 110` samples in total and return the last `N = 100` of them, so drop all warm-up samples.
  - If `num_warmup = 10`, `discard_initial = 20`, and `N = 100`, then we would sample `N + discard_initial = 120` samples in total and return the last `N = 100` of them, so we would drop all samples in the warm-up stage and the first 10 samples of the regular sampling.

torfjelde: Haha, aight. Well, the rest is _supposed_ to implement exactly what you outlined :sweat_smile: 

I'll see if I can also add some tests.

src/sample.jl
```diff
@@ -100,14 +122,21 @@ function mcmcsample(
...
+
+    # Determine how many samples to drop from `num_warmup` and the
+    # main sampling process before we start saving samples.
+    discard_from_warmup = min(num_warmup, discard_initial)
+    keep_from_warmup = num_warmup - discard_from_warmup
+    discard_from_sample = max(discard_initial - discard_from_warmup, 0)
```
devmotion, line 133: Yes, this seems to match what I wrote above.

src/sample.jl
torfjelde: Aaah I was totally unaware!

So I removed this, and then I've just added a section to `sample` to tell people where to find docs on the default arguments. I personally rarely go to the docs of a package unless I "have" to, so I think it's at least nice to tell the user where to find the info. I'm even partial to putting the stuff about common keywords in the _actual_ docstrings of `sample` but I'll leave as is for now.

src/sample.jl
```diff
@@ -124,34 +156,76 @@ function mcmcsample(
         # Obtain the initial sample and state.
         sample, state = step(rng, model, sampler; kwargs...)
```
devmotion, line 148: Shouldn't this use `step_warmup` as well, if `num_warmup > 0`? Or do we not want to make a difference between the stages regarding initial samples?

src/sample.jl
```diff
@@ -124,34 +156,76 @@ function mcmcsample(
...
+            callback === nothing ||
+                callback(rng, model, sampler, sample, state, i; kwargs...)
+
+            # Save the sample.
+            samples = AbstractMCMC.samples(sample, model, sampler; kwargs...)
+            samples = save!!(samples, sample, i, model, sampler; kwargs...)
+
+            # Increment iteration number.
+            i += 1
+        end
```
devmotion: Wouldn't it be easier to do something along the lines of
```julia
        # Discard initial samples, if needed.
        for _ in 1:discard_initial
            # Update the progress bar.
            if progress && i >= next_update
                ProgressLogging.@logprogress i / Ntotal
                next_update = i + threshold
            end

            # Obtain the next sample and state.
            sample, state = if i <= num_warmup
                step_warmup(rng, model, sampler, state; kwargs...)
            else
                step(rng, model, sampler, state; kwargs...)
            end
        end

        ...

        # Increment iteration number.
        i += 1
```

torfjelde: Well... Yes :facepalm:

src/sample.jl
torfjelde: I was thinking that the initial step shouldn't discriminate as to wether it's a warmup or not, but I guess it won't cost us anything + there might be a time where it makes sense, so let's do that :+1:

src/sample.jl
torfjelde: Ah actually, I now remember the reason for why I didn't do this: won't this make it type-unstable while the current implementation won't (if indeed `step_warmup` and `step` return the same type)?

Whether we care is another thing.

devmotion: If `step_warmup` and `step` return the same types, both approaches should be type-stable, no? And if not, you're in trouble in both cases I assume?

torfjelde: My worry was that too many if-staments involving runtime information could confuse the type-inference, even in the case where the return-types of each branch is the same, _but_ from your comments I'm assuming this was an irrational fear:)

Also, I just read a bit more https://juliahub.com/blog/2016/04/inference-convergence/ and my fears are indeed irrational :+1: I guess as long as each of the if-statement returns the same two types, i.e. the `Union` doesn't change, we're fine. IIRC Julia does union-splitting up to unions of length 4.

src/sample.jl
```diff
@@ -122,36 +144,62 @@ function mcmcsample(
...
 
         # Obtain the initial sample and state.
-        sample, state = step(rng, model, sampler; kwargs...)
+        sample, state = if num_warmup > 0
+            step_warmup(rng, model, sampler; kwargs...)
+        else
+            step(rng, model, sampler; kwargs...)
+        end
+
+        # Warmup sampling.
```
devmotion, line 153:
```suggestion
        # Discard initial samples.
```

src/sample.jl
```diff
@@ -122,36 +144,62 @@ function mcmcsample(
...
+        for j in 1:discard_initial
+            # Obtain the next sample and state.
```
devmotion, line 155: I think these should be accounted for in the progress logger as well (as done currently).

torfjelde: Should be good now :+1:

src/interface.jl
```diff
@@ -56,6 +56,21 @@ current `state` of the sampler.
...
 function step end
 
+"""
+    step_warmup(rng, model, sampler[, state; kwargs...])
+
+Return a 2-tuple of the next sample and the next state of the MCMC `sampler` for `model`.
+
+When sampling using [`sample`](@ref), this takes the place of [`step`](@ref) in the first
+`num_warmup` number of iterations, as specified by the `num_warmup` keyword to [`sample`](@ref).
+This is useful if the sampler has a "warmup"-stage initial stage that is different from the
```
devmotion, line 66:
```suggestion
This is useful if the sampler has an initial "warmup"-stage that is different from the
```

src/interface.jl
```diff
@@ -56,6 +56,21 @@ current `state` of the sampler.
...
+standard iteration.
+
+By default, this simply calls [`step`](@ref.)
```
devmotion, line 69:
```suggestion
By default, this simply calls [`AbstractMCMC.step`](@ref).
```

docs/src/design.md
```diff
@@ -63,6 +63,15 @@ the sampling step of the inference method.
 AbstractMCMC.step
 ```
 
+If one also has some special handling of the warmup-stage of sampling, then this can be specified by overloading
+
+```@docs
+AbstractMCMC.step_warmup
+```
+
+which will be used for the first `num_warmup`, as specified as a keyword argument to [`AbstractMCMC.sample`](@ref).
```
devmotion, line 72:
```suggestion
which will be used for the first `num_warmup` iterations, as specified as a keyword argument to [`AbstractMCMC.sample`](@ref). 
```

src/interface.jl
```diff
@@ -56,6 +56,21 @@ current `state` of the sampler.
 """
 function step end
 
+"""
+    step_warmup(rng, model, sampler[, state; kwargs...])
+
+Return a 2-tuple of the next sample and the next state of the MCMC `sampler` for `model`.
+
+When sampling using [`sample`](@ref), this takes the place of [`step`](@ref) in the first
```
devmotion, line 64: Maybe good to qualify the unexported functions?
```suggestion
When sampling using [`sample`](@ref), this takes the place of [`AbstractMCMC.step`](@ref) in the first
```

src/sample.jl
```diff
@@ -122,36 +144,68 @@ function mcmcsample(
...
+        else
+            step(rng, model, sampler; kwargs...)
+        end
+
+        # Update the progress bar.
+        itotal = 1
+        if progress && itotal >= next_update
+            ProgressLogging.@logprogress itotal / Ntotal
+            next_update = itotal + threshold
+        end
```
devmotion, line 158: I think this is not needed? At least it's not present on the master branch, it seems.
```suggestion
```

src/sample.jl
```diff
@@ -122,36 +144,68 @@ function mcmcsample(
...
+        samples = save!!(samples, sample, i, model, sampler; kwargs...)
 
-        # Update the progress bar.
-        itotal = 1 + discard_initial
-        if progress && itotal >= next_update
-            ProgressLogging.@logprogress itotal / Ntotal
-            next_update = itotal + threshold
+        # Step through remainder of warmup iterations and save.
+        i += 1
+        for _ in 1:keep_from_warmup
```
devmotion, line 188: Could we use the same approach here as well? And only add a `if i <= num_warmup` around the current `step` calls?

src/sample.jl
torfjelde: It's not present on the master branch because there we do progress-reporting in the beginning of the for-loop for the discarded samples, and then a final progress-report before we start the "main sampling loop". 

Here I've just moved the "final progress-report before we start the main sampling loop" to before all of the loops, and then moved the progres-reporting to the end of the for-loop for the discarded samples (well, I see I have forgotten to move the progress-reporting inside the discarded samples for-loop :grimacing: Will fix now)

src/sample.jl
torfjelde: We can do that :+1:

torfjelde: Ah, actually one thing: should we also do thinning for `num_warmup` then? :confused:

devmotion: I think there are different reasonable approaches but we could just keep doing what we've done so far: we discard `discard_initial` samples and then return the next `N` samples, thinned by a factor `thinning`. That is, the returned samples are samples in iteration `discard_initial + 1`, `discard_initial + thinning + 1`, `discard_initial + 2 * thinning + 1` etc.

torfjelde:
Would you mind having another look @devmotion ?:)

torfjelde:
Also, should I bump the minor version (currently bumped the patch version, which seems incorrect)?

src/sample.jl
```diff
@@ -122,40 +144,58 @@ function mcmcsample(
...
 
         # Discard initial samples.
-        for i in 1:discard_initial
-            # Update the progress bar.
-            if progress && i >= next_update
-                ProgressLogging.@logprogress i / Ntotal
-                next_update = i + threshold
+        for j in 1:discard_initial
+            # Obtain the next sample and state.
+            sample, state = if j ≤ num_warmup
```
devmotion, line 163: This should be
```suggestion
            sample, state = if j ≤ discard_num_warmup
```
shouldn't it?

Maybe it could even be split into two sequential for loops?

src/sample.jl
```diff
@@ -122,40 +144,58 @@ function mcmcsample(
...
+
         # Run callback.
-        callback === nothing || callback(rng, model, sampler, sample, state, 1; kwargs...)
+        callback === nothing || callback(rng, model, sampler, sample, state, i; kwargs...)
 
         # Save the sample.
-        samples = AbstractMCMC.samples(sample, model, sampler, N; kwargs...)
-        samples = save!!(samples, sample, 1, model, sampler, N; kwargs...)
+        samples = AbstractMCMC.samples(sample, model, sampler; kwargs...)
+        samples = save!!(samples, sample, i, model, sampler; kwargs...)
```
devmotion, line 184:
```suggestion
        samples = AbstractMCMC.samples(sample, model, sampler, N; kwargs...)
        samples = save!!(samples, sample, i, model, sampler, N; kwargs...)
```

src/sample.jl
```diff
@@ -122,40 +144,58 @@ function mcmcsample(
...
-        if progress && itotal >= next_update
-            ProgressLogging.@logprogress itotal / Ntotal
-            next_update = itotal + threshold
-        end
+        # Step through remainder of warmup iterations and save.
+        i += 1
 
         # Step through the sampler.
-        for i in 2:N
+        while i ≤ N
```
devmotion, line 190: Any particular reason to switch to a `while` loop here?

src/sample.jl
```diff
@@ -209,44 +256,65 @@ function mcmcsample(
...
+            step(rng, model, sampler; kwargs...)
+        end
 
-        # Discard initial samples.
-        for _ in 1:discard_initial
+        # Warmup sampling.
+        for j in 1:discard_initial
             # Obtain the next sample and state.
-            sample, state = step(rng, model, sampler, state; kwargs...)
+            sample, state = if j ≤ num_warmup
```
devmotion, line 284:
```suggestion
            sample, state = if j ≤ discard_num_warmup
```

src/sample.jl
```diff
@@ -209,44 +256,65 @@ function mcmcsample(
...
-        # Step through the sampler until stopping.
-        i = 2
+        samples = save!!(samples, sample, i, model, sampler; kwargs...)
 
         while !isdone(rng, model, sampler, samples, state, i; progress=progress, kwargs...)
             # Discard thinned samples.
             for _ in 1:(thinning - 1)
                 # Obtain the next sample and state.
-                sample, state = step(rng, model, sampler, state; kwargs...)
+                sample, state = if i ≤ keep_from_warmup
```
devmotion, line 305: Shouldn't `i` be incremented at the top of the loop? Before it was `2` here, now it is `1`.

devmotion (commented):
It seems tests are failing?

src/sample.jl
torfjelde: > This should be

I think _technically_ it doesn't matter, right? Since we have either
1. `discard_num_warmup = discard_initial` when `num_warmup >= discard_initial`, or
2. `discard_num_warmup = num_warmup` when `num_warmup < discard_initial`.

In both of those cases we get the same behavior in the above.

_But_ I think for readability's sake, I agree we should make the change! Just pointing out it shouldn't been a cause of a bug.

> Maybe it could even be split into two sequential for loops?

Wait what, wasn't that what I had before? :confused:

src/sample.jl
torfjelde: Ah no! I'll revert it to for-loop :+1:

src/sample.jl
torfjelde: Ah yes, nice catch!

src/sample.jl
```diff
@@ -209,44 +256,65 @@ function mcmcsample(
...
-        callback === nothing || callback(rng, model, sampler, sample, state, 1; kwargs...)
+        callback === nothing || callback(rng, model, sampler, sample, state, i; kwargs...)
 
         # Save the sample.
         samples = AbstractMCMC.samples(sample, model, sampler; kwargs...)
-        samples = save!!(samples, sample, 1, model, sampler; kwargs...)
-
-        # Step through the sampler until stopping.
-        i = 2
+        samples = save!!(samples, sample, i, model, sampler; kwargs...)
```
torfjelde, line 299:
```suggestion
        samples = save!!(samples, sample, i, model, sampler; kwargs...)
        
        i += 1
```

src/sample.jl
devmotion: > Wait what, wasn't that what I had before? confused

Really? I think you used a different logic initially but maybe I misremember :smile: In any case, I guess it does not matter.

torfjelde: You mean

```julia
for i = 1:discard_num_warmup
    # ...
end

for i = discard_num_warmup + 1:discard_initial
    # ...
end
```

?

Because you're probably right, I don't think I ever did this exactly :grimacing: 

I'm preferential to the current code for readability's sake because it means the discard stepping is looks the same as the proper stepping, code-wise.

devmotion:
@torfjelde Can you fix the merge conflict?

torfjelde:
Should be good to go now @devmotion

torfjelde:
But we're having issues with nightly on x86 so these two last ones won't finish.

src/sample.jl
```diff
@@ -179,6 +221,9 @@ function mcmcsample(
             # Save the sample.
             samples = save!!(samples, sample, i, model, sampler, N; kwargs...)
 
+            # Increment iteration counter.
+            i += 1
+
```
devmotion, line 226: This is handled by the `for i in 2:N` above:
```suggestion
```

src/sample.jl
```diff
@@ -214,28 +259,45 @@ function mcmcsample(
...
+                step(rng, model, sampler, initial_state; kwargs...)
+            end
         end
 
         # Discard initial samples.
-        for _ in 1:discard_initial
+        for j in 1:discard_initial
             # Obtain the next sample and state.
-            sample, state = step(rng, model, sampler, state; kwargs...)
+            sample, state = if j ≤ discard_from_warmup
```
devmotion, line 296: A bit inconsistent, you used
```suggestion
            sample, state = if j ≤ num_warmup
```
above. Doesn't matter (both will work) but I think it would be good to be consistent.

src/sample.jl
```diff
@@ -214,28 +259,45 @@ function mcmcsample(
...
     progressname="Convergence sampling",
     callback=nothing,
-    discard_initial=0,
+    num_warmup=0,
+    discard_initial=num_warmup,
     thinning=1,
     initial_state=nothing,
     kwargs...,
 )
+    # Determine how many samples to drop from `num_warmup` and the
```
devmotion, line 268: Can you add the same/similar error checks as above?

src/sample.jl
```diff
@@ -247,16 +309,23 @@ function mcmcsample(
...
+                sample, state = if i ≤ keep_from_warmup
+                    step_warmup(rng, model, sampler, state; kwargs...)
+                else
+                    step(rng, model, sampler, state; kwargs...)
+                end
             end
 
             # Obtain the next sample and state.
-            sample, state = step(rng, model, sampler, state; kwargs...)
+            sample, state = if i ≤ keep_from_warmup
```
devmotion, line 324: We could merge this with the for-loop above AFAICT?

torfjelde:
Ran into another place where this will be useful: https://github.com/torfjelde/AutomaticMALA.jl. Avoids having to put `num_adapts` in the sampler itself, and can be replaced with `num_warmup` as a kwarg.

(I need to do some more work on this PR; had sort of forgotten about this)

src/sample.jl
torfjelde: Aye, but we can do that everywhere here no? I can make this change, but I'll wait until you've had a final look (to make the diff clearer).

src/sample.jl
torfjelde: Done :+1:

torfjelde:
Aight @devmotion I believe I've addressed your comments:) Should be a quick approve now I think :+1: 

Sorry it took so long :grimacing:

torfjelde:
Bumped minor version btw, given that this adds a new feature.

devmotion (approved):

PR #119, opened 2023-03-13
Allow specification of initial state for `sample`

torfjelde:
This seems convenient to have, e.g. for resuming sampling, running special warm-up procedures.

EDIT: Note that this is now dependent on #126


src/sample.jl
```diff
@@ -466,11 +489,9 @@ function mcmcsample(
...
                         return chain
                     end
-                    chains = if init_params === nothing
-                        Distributed.pmap(sample_chain, pool, seeds)
-                    else
-                        Distributed.pmap(sample_chain, pool, seeds, init_params)
-                    end
+                    chains = Distributed.pmap(
+                        sample_chain, pool, seeds, _init_params, _initial_state
+                    )
```
devmotion, line 494: Does this work? I don't think `pmap` broadcasts its arguments?

devmotion (commented):
1. Can you add tests of all new lines?
2. The names `init_params` and `initial_state` seem inconsistent, can we either use `init` or `initial` in both cases?

src/sample.jl
torfjelde: It doesn't have to, right? `_init_params` will either be `fill(nothing, nchains)` or it will be `_initial_state` which should also be a vector of the correct length (I should add a check for this though)

torfjelde:
> Can you add tests of all new lines?

Will do (but tomorrow; am about to sleep)!

> The names init_params and initial_state seem inconsistent, can we either use init or initial in both cases? 

Agreed. Hmm, I guess I'll make it `init_state` then? Personally prefer the `initial_*` but given that one already exists, I guess it makes sense to add this (though this `init_params` isn't even offiically supported, right?)

src/sample.jl
devmotion: Ah yeah, you created these arrays. The motivation for the branch here was to avoid allocating such arrays.

devmotion:
> though this init_params isn't even offiically supported, right?

It's officially supported: It's clearly documented and used in downstream packages such as EllipticalSliceSampling, DynamicPPL, and Turing.

src/sample.jl
torfjelde: Yeah, am aware :+1: But the branching becomes a bit more annoying if we also have to do this for `init_state`, so I figured just allocating was a better. I can make it a conditional if you prefer :+1:

torfjelde:
> It's officially supported: It's clearly documented and used in downstream packages such as EllipticalSliceSampling, DynamicPPL, and Turing.


Yes, that I'm fully aware of! Just remembered seeing the following in the docs:

> There is no "official" way for providing initial parameter values yet. However, multiple packages such as [EllipticalSliceSampling.jl](https://github.com/TuringLang/EllipticalSliceSampling.jl) and [AdvancedMH.jl](https://github.com/TuringLang/AdvancedMH.jl) support an init_params keyword argument for setting the initial values when sampling a single chain. To ensure that sampling multiple chains "just works" when sampling of a single chain is implemented, [we decided to support init_params in the default implementations of the ensemble methods](https://github.com/TuringLang/AbstractMCMC.jl/pull/94):

Which I took to mean "we haven't really made an explicit decision on how to support initial parameters, but because so many downstream packages use `init_params`, we stay compatible with it". 

But nonetheless, I'll change it to `init_state` then :+1:

devmotion:
Generally, I think I'd prefer `initial` but yeah, it involves more changes :shrug: I'm also not a big fan of the name `params`, maybe `initial_sample` would be better?

Since #120 requires a breaking release anyway, we could also include more breaking changes. Otherwise we could deprecate the keyword argument, maybe something like the following could work:
```julia
function f(...; init_params=nothing, initial_params=init_params)
    if init_params !== nothing
        if initial_params !== init_params
            throw(ArgumentError("..."))
        end
        Base.deprecate("....", f)
    end
    ...
end
```

torfjelde:
@devmotion I just came across this again; given that we decided to scrap #120 in the end, what do you think of at least merging this? Being able to specify the initial state would be quite useful.

torfjelde:
I decided to just rip the band-aid off and go with renaming `init_params` to `initial_params` right away.

If we don't, we end up in a somewhat awkward scenario where we have to pass along both `init_params` and `initial_params`, which might also just break downstream `step`.

devmotion:
Can we merge #126 first and then rebase this PR on it? Or maybe even directly rebase the PR on #126 and adjust the base branch of the PR?

torfjelde:
Rebased and change base for PR.

I'll add some tests for the `intial_state` stuff sometime today :+1:

devmotion:
Seems the PR breaks EllipticalSliceSampling? 🤔 Interesting, even though it won't cause any breakage in practice if we tag a breaking release.

torfjelde:
Could it maybe be something related to `init_params` -> `initial_params`?

But yes, I'll bump the major version so won't break anything in practice:)

devmotion:
Yeah, I assume that's the reason: https://github.com/TuringLang/EllipticalSliceSampling.jl/blob/ca4babb2baba9008805bc8234a6fd182119e57dc/src/abstractmcmc.jl#L25 https://github.com/TuringLang/EllipticalSliceSampling.jl/blob/ca4babb2baba9008805bc8234a6fd182119e57dc/test/simple.jl#L40 I wonder though why other packages that currently use `init_params` (AdvancedMH IIRC? Turing?) are not affected by this change. Maybe they are missing tests for `init_params`/`initial_params`?

src/sample.jl
```diff
@@ -410,8 +428,14 @@ function mcmcsample(
...
-    # Ensure that initial parameters are `nothing` or of the correct length
-    check_initial_params(init_params, nchains)
+    # Ensure that initial parameters and states are `nothing` or of the correct length
+    check_initial_params(initial_params, nchains)
+    check_initial_state(initial_state, nchains)
+
+    _initial_params =
+        initial_params === nothing ? FillArrays.Fill(nothing, nchains) : initial_params
+    _initial_state =
+        initial_state === nothing ? FillArrays.Fill(nothing, nchains) : initial_state
```
devmotion, line 438: I think it would be nice to avoid these `FillArray`s. Maybe we could
- move the function below to a callable struct (could be shared between all ensemble algorithms maybe?)
- pass `initial_params` and `initial_state` to the constructor as well but only use it to define type parameters that allow us to distinguish between the four possible cases (no initial params and no initial state, only initial state, only initial params, and both initial params and state)
- define the function of the callable struct depending on the type parameters, forwarding the versions with only the seed or only the seed and one additional argument to the three-argument version

torfjelde:
> Maybe they are missing tests for init_params/initial_params?

Very likely :confused:

src/sample.jl
torfjelde: Is adding in callable structs here really an improvement? :confused: I agree it's more efficient, but it seems like this will be quite a bit more complex + the efficiency doesn't really matter here, right?

devmotion: A callable struct should generally be better for the compiler, shouldn't it? Regardless of whether we change or add arguments as in this PR.

torfjelde: Yep! But is this performance critical code? And it seems to be me that we'll need a callable struct for each scenario?

devmotion: 🤷 Are you sure? To me it seems one struct is sufficient - both the multithreaded and the multicore version seem to use the same inner structure, and in the serial case we could set `channel = nothing`. If needed we could also dispatch on the type of the algorithm to handle minor differences in the function call.

torfjelde: > To me it seems one struct is sufficient

To clarify, I don't question whether we can have a single callable struct with different call implementations; I meant more that it seems it won't be as simple as just doing

``` julia
struct SampleFunc
    # ...
end

function (f::SampleFunc)(args...)
    # ...
end
```

> multithreaded and the multicore version seem to use the same inner structure

But if we put `initial_params` and `initial_state` in the callable struct, then we'll need to `pmap`, etc. over a range containing the corresponding indices, no? Which seems like it would lead to more allocations than the current impl using `Fill(nothing, nchains)`?

Or am I misunderstanding what you mean here?

torfjelde: Just bumping this discussion:) Would be nice to get a version of this PR merged.

devmotion (approved):
Let's go with the FillArrays dependency for now.

Issue #118, opened 2023-03-10
Default `bundle_samples` is quite annoying


torfjelde:
The default implementation of `bundle_samples` specializes on `Vector{T}`, which in becomes very annoying if you want to implement your own `bundle_samples`, e.g. if you then use `AbstractVector` you almost immediately run into ambiguity errors.

```julia
https://github.com/TuringLang/AbstractMCMC.jl/blob/7192263f967b4a6bb1f78957c734a6a41c383026/src/interface.jl#L38-L44
```

In addition, IIUC it seems quite "useless" given that by default `sample` already returns a `Vector` of the transitions?

Thoughts on removing this?

torfjelde:
Do you have any thoughts on this @devmotion ?

PR #120, opened 2023-03-13
Remove overly specialized bundle_samples

torfjelde:
Related: #118

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "4.4.0"
+version = "4.5.0"
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -35,14 +35,6 @@ function bundle_samples(
     return samples
 end
 
-function bundle_samples(
-    samples::Vector, ::AbstractModel, ::AbstractSampler, ::Any, ::Type{Vector{T}}; kwargs...
-) where {T}
-    return map(samples) do sample
-        convert(T, sample)
-    end
-end
-
 """
     step(rng, model, sampler[, state; kwargs...])
 
--- a/test/sample.jl
+++ b/test/sample.jl
@@ -553,21 +553,6 @@
         @test all(chain[i].b == ref_chain[i].b for i in 1:N)
     end
 
-    @testset "Sample vector of `NamedTuple`s" begin
-        chain = sample(MyModel(), MySampler(), 1_000; chain_type=Vector{NamedTuple})
-        # Check output type
-        @test chain isa Vector{<:NamedTuple}
-        @test length(chain) == 1_000
-        @test all(keys(x) == (:a, :b) for x in chain)
-
-        # Check some statistical properties
-        @test ismissing(chain[1].a)
-        @test mean(x.a for x in view(chain, 2:1_000)) ≈ 0.5 atol = 6e-2
-        @test var(x.a for x in view(chain, 2:1_000)) ≈ 1 / 12 atol = 1e-2
-        @test mean(x.b for x in chain) ≈ 0 atol = 0.1
-        @test var(x.b for x in chain) ≈ 1 atol = 0.15
-    end
-
     @testset "Testing callbacks" begin
         function count_iterations(
             rng, model, sampler, sample, state, i; iter_array, kwargs...
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -79,9 +79,6 @@ function AbstractMCMC.chainscat(
     return vcat(chain, chains...)
 end
 
-# Conversion to NamedTuple
-Base.convert(::Type{NamedTuple}, x::MySample) = (a=x.a, b=x.b)
-
 # Gaussian log density (without additive constants)
 # Without LogDensityProblems.jl interface
 mylogdensity(x) = -sum(abs2, x) / 2


devmotion (commented):
Do you have a concrete example that is fixed by this removal? Removing this definitions seems very breaking. Without this method the return types will be incorrect, won't they? The main point is that the elements are converted to type `T` which won't happen anymore.

torfjelde:
> Do you have a concrete example that is fixed by this removal? Removing this definitions seems very breaking. Without this method the return types will be incorrect, won't they? The main point is that the elements are converted to type T which won't happen anymore.

Nothing is _fixed_, it's just that
1. I don't actually know of anywhere this is being used. Do we know of any place where this is the case?
2. If you want to do something with `Type{Vector{...}}` then you need to override this default implementation. This sort of thing comes up very often once you have samplers which are working on a "multiple" models, e.g. if you implement tempering. In the case it seems very reasonable to allow `Vector{MCMCChains.Chains}` for `samples::AbstractVector`, which then requires explicit overriding of this default implementation.
3. `map` + `convert` is _very_ easy to do for the user themselves, to the point where it doesn't seem worth the hassle it causes :shrug:

torfjelde:
> Without this method the return types will be incorrect, won't they? The main point is that the elements are converted to type T which won't happen anymore.

And this is not even a promise we are making, right? For example, `bundle_samples` will currently just be identity if you provide it with a type it has no implementation for (this is a different question: should we error here?)

torfjelde:
An alternative is of course to introudce a

```julia
AbstractMCMC.bundle_samples_to_vector
```

or something, i.e. just an additional diversion so that overloading for `samples::AbstractVector` doesn't become such a hassle, which is then the default (and thus keeps the change non-breaking).

And let me try to clarify a bit the issue I have with the current approach.

If you want to overload `AbstractMCMC.bundle_samples(::AbstractVector, ..., ::Type{T})` where `T`, and, say, just make some transformation to the `samples` before passing them on to the "actual" `bundle_samples` implementation (this is very common in the case of "meta-samplers", e.g. tempered samplers, composition of samplers), then you have to do the following

```julia
function AbstractMCMC.bundle_samples(samples, ..., ::Type{T}) where {T}
    return actual_bundle_samples(samples, ..., T)
end

function AbstractMCMC.bundle_samples(samples::Vector, ..., ::Type{Vector{T}}) where {T}
    return actual_bundle_samples(samples, ..., Vector{T})
end
```

which gets quite annoying.

devmotion:
I've been annoyed by downstream issues/annoyances of the `bundle_samples` interface as well. The reason why I'm a bit skeptic about the PR in its current form is that it is breaking (clearly indicated by the broken tests) and hence should be released in a breaking release, regardless of how useful the change is. But when we go through the hassle of updating all downstream packages anyway, I wonder if we should spend a few days on thinking about whether we could/should improve the interface more generally. For instance, currently it's not intended by the interface to dispatch on the output-type in the `sample` call - is that an issue? Is the output type useful at all? Should we remove `bundle_samples` completely, what would the implications be? Could/should we generally reduce the number of arguments in the interface API?

torfjelde:
Oooh AbstractMCMC is already on x.y.z versioning! I was thinking bumping minor version was a breaking release, and so was confused why you were talking about how this should be a breaking release thinking I'd already indicated so.

> But when we go through the hassle of updating all downstream packages anyway, I wonder if we should spend a few days on thinking about whether we could/should improve the interface more generally. For instance, currently it's not intended by the interface to dispatch on the output-type in the sample call - is that an issue? Is the output type useful at all? Should we remove bundle_samples completely, what would the implications be? Could/should we generally reduce the number of arguments in the interface API?

And yeah, happy to spend a few days thinking aobut whether or not this is the way to go :+1: 

Should we maybe make an issue or discussion?

torfjelde:
> currently it's not intended by the interface to dispatch on the output-type in the sample call - is that an issue? 

> Is the output type useful at all? 

When you refer to the "output type in the sample call", are you talking about the `samples` argument of `bundle_samples` or the output-type specified to `bundle_samples`? I'm guessing the latter, but just asking to be sure.

And btw, a lot of these PRs are motivated by "annoyances" I've encountered when working on MCMCTempering.jl, where we have several "meta-samplers", i.e. samplers which call other sampler in some way.

In MCMCTempering.jl we have a `MultiModel`, which effectively defines a product of models but with an ordering (for implementation purposes only). In this case, it might be sensible for the user to ask for a `Vector{MCMCChains.Chains}`, with each chain corresponding to each model.

Another particularly "weird" case where `bundle_samples` is used (though not related to the dispatching on the output-type), though IMO it also seems like a decent way to handle this stuff this method: https://github.com/TuringLang/MCMCTempering.jl/blob/ed1ca9886d1c49aece23aacf2bc9fcaf77725fd5/src/MCMCTempering.jl#L93-L108

In this case, the samples correspond to samples from a product of models, and so it represents an ordered sequence of samples rather than a single one. _But_ the sequence is ordered according to an "index process", which is not necessarily equal to the model-ordering (which is what you'd expect as a user). To "handle" this in a somewhat more convenient manner for the user, we've added a kwarg to `bundle_samples` called `bundle_resolve_swaps` which will automatically re-order the sequence in each sample to align with the provided models.

> Should we remove bundle_samples completely, what would the implications be?

Do you mean completely removing it and just not doing anything with the samples, leaving that to the sampler-implementer? If so, I'm not a big fan, in particular not now that we have extensions and people can easily add impls for MCMCChains, etc.

> Could/should we generally reduce the number of arguments in the interface API?

What we could always do, though it has to be done with care, is to add some simpler `bundle_samples` method which can be used if all the sampler-information isn't useful, i.e.

```julia
function bundle_samples(
    samples, ::AbstractModel, ::AbstractSampler, ::Any, output_type::Type; kwargs...
)
    return bundle_samples(samples, output_type; kwargs...)
end
```

and then just make the default implementation we complained about above

```julia
function bundle_samples(samples::Vector, ::Type{Vector{T}}; kwargs...) where {T}
    return map(samples) do sample
        convert(T, sample)
    end
end
```

I believe in _most_ (but not all) scenarions, one would really only overload the latter implementation.

torfjelde:
@devmotion Could we revisit this?:)

devmotion:
I'm still unhappy. The least this should be moved to a breaking 5.0.0 release.

However, I wonder - could we just move the functionality to the body of the generic fallback, wrapped in a `if samples isa Vector && T <: Vector` (or similar) check to fix the method ambiguity issues without actually removing functionality?

devmotion:
Since #123 was merged, I suggest closing this PR for now and revisit it (or some variant of it) if these ambiguity issues show up again.

torfjelde:
Sounds good :+1:

PR #123, opened 2023-06-17
Fix method ambiguity issues

devmotion:
Alternative to #120 which keeps the current functionality.

@torfjelde Do you have a MWE of the ambiguity issues? Or can confirm that this PR would fix the issues you ran into?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "4.4.1"
+version = "4.4.2"
 
 [deps]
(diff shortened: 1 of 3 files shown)


torfjelde:
I'll have to dig out those examples; it's been a while. But will try to get around to it sometime today.

devmotion:
Were you able to find (and run?) the examples?

torfjelde:
No, sorry haven't done it yet. The original examples in #120 are from MCMCTempering.jl which I haven't done in a while. This PR most certainly helps with the overall issue we've discussed, whether or not it actuallly solves the particular ambiguity I ran into.

devmotion:
Should we merge then maybe merge the PR and revisit #120 (or some variant) if problems show up again?

torfjelde:
I'm happy with that:) I have NeurIPS reviews over the next few weeks, but once that's done I'll revisit MCMCTempering, and so might be able to provide some cases in #120 then :+1: But for now, let's just merge this.

torfjelde (approved):
Thanks David:)

PR #124, opened 2023-07-29
Default`n_adapts` and `discard_initial` to zero 

yebai:
Some downstream MCMC sampling package, e.g., [AHMC](https://github.com/TuringLang/Turing.jl/pull/2050#discussion_r1278265961), now assumes we always pass `n_adapts` to the `AbstractMCMC.step` function. It might be sensible to default these arguments to 0 when users didn't specify them to avoid missing `n_adapts` errors.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "4.4.2"
+version = "4.4.3"
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"


src/sample.jl
```diff
@@ -100,6 +100,7 @@ function mcmcsample(
     progress=PROGRESS[],
     progressname="Sampling",
     callback=nothing,
+    n_adapts=0,
```
devmotion, line 103: `n_adapts` is not a general thing (in contrast to `discard_initial`), so my feeling is this should not be specified in AbstractMCMC but downstream packages where it is relevant.

src/sample.jl
```diff
@@ -395,6 +399,8 @@ function mcmcsample(
     ::MCMCDistributed,
     N::Integer,
     nchains::Integer;
+    n_adapts=0,
+    discard_initial=0,
```
devmotion, line 403: This should not be needed - the keyword arguments are forwarded to the single-chain version and there a default of 0 is already specified.

devmotion: The same is true for the other ensemble algorithms.

Issue #125, opened 2023-09-13
Error if `init_params` has not  been specified for all chains 


torfjelde:
It seems we don't error correctly for certain multi-chain sample calls; in particular, `Serial` and `MCMCDistributed` seem to have this issue

https://github.com/TuringLang/AbstractMCMC.jl/blob/d7c549fe41a80c1f164423c7ac458425535f624b/src/sample.jl#L522-L526

https://github.com/TuringLang/AbstractMCMC.jl/blob/d7c549fe41a80c1f164423c7ac458425535f624b/src/sample.jl#L469-L473

while `MCMCThreads` does not, as it makes of `_first_or_nothing`:

https://github.com/TuringLang/AbstractMCMC.jl/blob/d7c549fe41a80c1f164423c7ac458425535f624b/src/sample.jl#L315C21-L316

https://github.com/TuringLang/AbstractMCMC.jl/blob/d7c549fe41a80c1f164423c7ac458425535f624b/src/sample.jl#L542-L549

Doesn't seem like there's a reason why we don't use `_first_or_nothing` for `MCMCSerial` and `MCMCDistributed`; guessing it's just an oversight?

Ref: https://github.com/TuringLang/Turing.jl/issues/2079

torfjelde:
Note that the check in `_first_or_nothing` is not sufficient to catch scenarios where `length(y) == n` just happens to hold because the number of chains is the same as the dimension of the parameters :grimacing: But we can't really catch this here given that elements of `init_params` could be anything (even if `init_params[1]` is a `Float64`, it could be that the inner `sample` call expects `Float64` as the `init_params`  argument)

devmotion:
> Doesn't seem like there's a reason why we don't use _first_or_nothing for MCMCSerial and MCMCDistributed

Can you explain why we should use it? In those two cases we call `map`/`pmap` which require argument of matching length already, so it should not be possible to use `init_params` of incorrect length. We also can't call these functions with `nothing`, so we have to check for `nothing` explicitly anyway and then call the method without additional argument.

torfjelde:
> we call map/pmap which require argument of matching length already

It doesn't :grimacing: `map` will simply just use the shortest inputs:

``` julia
julia> map(+, 1:2, 3:5)
2-element Vector{Int64}:
 4
 6
```

Alternatively we can use `zip` or something, which does require same length inputs

devmotion:
[...]

Issue #109, opened 2022-11-17
`sample` which takes in an initial state


torfjelde:
It would be nice to move have something similar to `resume_from`  that we have in DPPL (https://github.com/TuringLang/DynamicPPL.jl/blob/217b8cc3a0fbee96049b90a53f5db9f2a586ba96/src/sampler.jl#L84-L95) in AbstractMCMC.

In particular, by default, it shouldn't be an issue to just make
```julia
loadstate(state) = state
```
assuming that the `resume_from` corresponds to the sampler `state`.

Thoughts @devmotion @cpfiffer @yebai ?

CC: @HarrisonWilde

yebai:
I think it is a useful feature for all samplers, which would fit nicely here.

yebai:
To slightly widen the scope of this issue, we can also consider saving/restoring sampler states (e.g. step-size and preconditioning matrix of HMC).

torfjelde:
> To slightly widen the scope of this issue, we can also consider saving/restoring sampler states (e.g. step-size and preconditioning matrix of HMC).

Saving in what sense? This would all be included in the `state`, no?

devmotion:
Yes, I think that's one of the things that could/should be moved from DynamicPPL/Turing to AbstractMCMC (maybe I misremember but it feels like there was also some discussion in some issue or PR a while ago?).

torfjelde:
Closed by https://github.com/TuringLang/AbstractMCMC.jl/pull/119

PR #126, opened 2023-09-13
Use _init_parmas for MCMCThreads and MCMCDistributed too

torfjelde:
Fixes https://github.com/TuringLang/AbstractMCMC.jl/issues/125

### DIFF
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -544,7 +544,7 @@ check_initial_params(x::Nothing, n::Int) = nothing
 function check_initial_params(x, n::Int)
     if length(x) != n
         throw(
-            ArgumentError("not enough initial parameters (expected $n, received $(length(x))"),
+            ArgumentError("incorrect number of initial parameters (expected $n, received $(length(x))"),
         )
     end
 end
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -314,8 +314,6 @@ function mcmcsample(
 
     # Ensure that initial parameters are `nothing` or of the correct length
     check_initial_params(init_params, nchains)
-    # We will use `getindex` later so we need to `collect`.
-    _init_params = init_params !== nothing ? collect(init_params) : nothing
 
     # Set up a chains vector.
     chains = Vector{Any}(undef, nchains)
@@ -366,10 +364,10 @@ function mcmcsample(
                                 _sampler,
                                 N;
                                 progress=false,
-                                init_params=if _init_params === nothing
+                                init_params=if init_params === nothing
                                     nothing
                                 else
-                                    _init_params[chainidx]
+                                    init_params[chainidx]
                                 end,
                                 kwargs...,
                             )
@@ -540,8 +538,14 @@ end
 tighten_eltype(x) = x
 tighten_eltype(x::Vector{Any}) = map(identity, x)
 
-check_initial_params(x::Nothing, n::Int) = nothing
-function check_initial_params(x, n::Int)
+@nospecialize check_initial_params(x, n) = throw(
+    ArgumentError(
+        "initial parameters must be specified as a vector of length equal to the number of chains or `nothing`",
+    ),
+)
+
+check_initial_params(::Nothing, n) = nothing
+function check_initial_params(x::AbstractArray, n)
     if length(x) != n
         throw(
             ArgumentError(
@@ -549,4 +553,6 @@ function check_initial_params(x, n::Int)
             ),
         )
     end
+
+    return nothing
 end
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -314,6 +314,8 @@ function mcmcsample(
 
     # Ensure that initial parameters are `nothing` or of the correct length
     check_initial_params(init_params, nchains)
+    # We will use `getindex` later so we need to `collect`.
+    _init_params = collect(init_params)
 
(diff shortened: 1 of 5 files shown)


test/sample.jl
```diff
@@ -184,14 +185,36 @@
...
+
+        # Too many `init_params`
+        @test_throws ArgumentError sample(
+            MyModel(),
+            MySampler(),
+            MCMCThreads(),
+            3,
+            nchains;
+            progress=false,
+            init_params=Iterators.repeated(init_params, nchains + 1),
```
torfjelde, line 205: So `_first_or_nothing` doesn't actually care if there are too many initial parameters :confused: 

IMO it seems like we want to raise an error in this case since it's otherwise very easy to accidentally do the incorrect thing.

torfjelde:
So I just removed `_first_or_nothing` and added a simple check for the length of `init_params`.

Though this is then a breaking change

src/sample.jl
```diff
@@ -312,8 +312,10 @@ function mcmcsample(
     # Create a seed for each chain using the provided random number generator.
     seeds = rand(rng, UInt, nchains)
 
-    # Ensure that initial parameters are `nothing` or indexable
-    _init_params = _first_or_nothing(init_params, nchains)
+    # Ensure that initial parameters are `nothing` or of the correct length
+    check_initial_params(init_params, nchains)
+    # We will use `getindex` later so we need to `collect`.
+    _init_params = init_params !== nothing ? collect(init_params) : nothing
```
devmotion, line 318: Can we remove the `collect` call? It's not generally required and will lead to surprising allocations, e.g., the following cases would work just fine without `collect` (IIRC the docs even mention FillArrays if you want to specify the same parameters multiple times):
```julia
julia> collect(1:10)
10-element Vector{Int64}:
  1
  2
  3
  4
  5
  6
  7
  8
  9
 10

julia> collect(Zeros(10))
10-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
```

```suggestion
```

torfjelde: So I added it explicitly to cause allocations, but maybe it's not quite the right thing to do. We use `getindex` further down, and so if `init_params` is an iterator, e.g. `Iterator.repeated`, then this will fail unless we do something like `collect` first.

torfjelde: How about we just simplify the process by adding something like:

``` julia
multiple_init_params(::Nothing, n) = Iterator.repeated(nothing, n)
function multiple_init_params(x, n)
    if length(x) != n
        throw(
            ArgumentError(
                "incorrect number of initial parameters (expected $n, received $(length(x))"
            ),
        )
    end

    return x
end
```

and then we just do

``` julia
_init_params = multiple_init_params(init_params, n)
```

then we can just remove all the subsequent checks for `nothing` + use `zip` and avoid the `getindex` call.

devmotion: I think we should not generally change `nothing` to `Iterator.repeated`. Mainly, because it means we would not dispatch on `nothing` anymore and hence not use the `map`/`pmap` calls with reduced number of arguments anymore. IMO we really want to use these calls with one argument less since generally `map`, `broadcast` etc. is more optimized for `AbstractArray` than generic iterators.

For this reason I'm also not a fan of promoting and using `Iterators.repeated`. Maybe we should just demand that multiple sets of parameters are always provided as an `AbstractVector`? If you want something cheap, you can just use e.g. `Fill(...)` instead of `Iterators.repeated` - which in contrast to the latter supports the array interface.

torfjelde: >  IMO we really want to use these calls with one argument less since generally map, broadcast etc. is more optimized for AbstractArray than generic iterators.

Good point!

> If you want something cheap, you can just use e.g. Fill(...) instead of Iterators.repeated - which in contrast to the latter supports the array interface.

Am happy with this :+1:

devmotion: Could you update the docs (https://turinglang.org/AbstractMCMC.jl/dev/api/#Common-keyword-arguments) accordingly and remove `Iterators.repeated` there?

torfjelde: Will do :+1:

torfjelde:
@devmotion given that we're making a breaking change here, should we also throw in the renaming of `init_params` to `initial_params` as discussed in #109 ?

devmotion (commented):
Looks good. Adding `using FillArrays: FillArrays` to the tests should fix the test errors.

devmotion (approved):

torfjelde:
Aight, I'll merge this now, but let's delay the release until #119 has gone through to limit the number of breaking releases :+1:

devmotion:
But the PR here was marked as non-breaking anyway? 🤔 (I saw you already corrected the version bump 👍)

torfjelde:
Wait, isn't minor version breaking?

And this will be breaking, since we no longer accept previously accepted
input types.

devmotion:
No, minor versions are only breaking < 1.0.0. The next breaking release will be 5.0.0.

torfjelde:
Ooooooh, crap. I clearly work too much with < 1.0.0 packages :grimacing: 

But aight, I'll bump to 5.0.0 then?

devmotion:
👍

PR #133, opened 2023-10-24
Reversion of commits v4.4.2..v4.5.0

torfjelde:
This supersedes #132, as in that PR I didn't do a proper `git revert` of all the commits.

@yebai @devmotion

PR #134, opened 2023-10-24
Breaking release: v5.0.0

torfjelde:
Follow-up from #133.


devmotion (approved):
Looks good 🙂 (The git history is surprisingly complicated for this change 😅)

torfjelde:
Ooooh true :facepalm: I was planning on applying this after #133 . I'll fix that.

torfjelde:
Seems like something strange is going on with x86 on nightly? It's been running for ages.

Also, is there a reason why we've running CI with some many targets? Seems somewhat wasteful for something as "simple" as AbstractMCMC?

PR #138, opened 2024-02-26
Add option to mute the logging for `setprogress!` function

sunxd3:


Project.toml
```diff
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "5.1.0"
+version = "5.1.1"
```
devmotion, line 6: Following semver, I think we should change this to
```suggestion
version = "5.2.0"
```
since the PR is not a bugfix but a new feature.

src/sample.jl
```diff
@@ -6,8 +6,10 @@ const PROGRESS = Ref(true)
 
 Enable progress logging globally if `progress` is `true`, and disable it otherwise.
 """
-function setprogress!(progress::Bool)
-    @info "progress logging is $(progress ? "enabled" : "disabled") globally"
+function setprogress!(progress::Bool; verbose::Bool=true)
```
devmotion, line 9: I'm always a bit unsure if it's more natural if the default value of a boolean flag is `false` or `true`. In https://www.serendipidata.com/posts/naming-guidelines-for-boolean-variables a few arguments are provided for why one might prefer
``suggestion
function setprogress!(progress::Bool; silent::Bool=false)
```

devmotion:
@sunxd3 can you open a follow-up PR that fixes the documentation?

PR #144, opened 2024-07-12
Add some interface functions to support the new Gibbs sampler in Turing

sunxd3:
The recent [new Gibbs sampler](https://github.com/TuringLang/Turing.jl/pull/2231) provides a way forward for the Turing inference stack. 

A near-to-medium-range goal has been to further reduce the glue code between Turing and inference packages (ref https://github.com/TuringLang/Turing.jl/issues/2281). The new Gibbs implementation laid a great plan to achieve this goal.

This PR is modeled after the interface of @torfjelde's [recent PR](https://github.com/TuringLang/Turing.jl/blob/torfjelde/gibbs-new-improv/src/mcmc/abstractmcmc.jl). And in some aspects, it is a rehash of https://github.com/TuringLang/AbstractMCMC.jl/pull/86. 

(the explanation here is outdated, please refer to https://github.com/TuringLang/AbstractMCMC.jl/pull/144#issuecomment-2337681868)

~~The goal of this PR is to determine and implement some **necessary** interface improvements, so that, when we update the inference packages up to the interface, they will more or less "just work" with the new Gibbs implementation.~~

~~As a first step, we test-flight two new functions `recompute_logprob!!(rng, model, sampler, state)` and `getparams(state)`:~~
- ~~`recompute_logprob!!(rng, model, sampler, state)` recomputes the logprob given the `state`~~
- ~~`getparams(state)` extract the parameter values~~

~~Some considerations:~~
- ~~This assumes a `state` is implemented with `AbstractMCMC` compatible inference packages. And a `state` at least stores values of parameters from the current iteration (traditionally, this is in the form of a `Transition`) and logprob.~~
- ~~`recompute_logprob!!(rng, model, sampler, state)`~~
    - ~~do we need `rng`?~~
    - ~~should we make `model` into `AbstractMCMC.LogDensityModel` or just `LogDensityProblem` (and make inference packages depend on `LogDensityProblems` in the latter case)? This should allow inference packages to be independent from DynamicPPL, we can use `getparams` to construct a `varinfo` in Turing~~
- ~~`getparams(state)` ~~
    - ~~What does this function return? A vector, a `transition`?~~
    - ~~Do we need `setparams`?~~
- ~~Do we also need some interface functions for `state` like `getstats`?~~

~~Tor also says (in a Slack conversation) that the a `condition(model, params)` is needed, but better to be implemented by packages that defines the `model`, which I agree.~~


devmotion:
How is #86 related to this PR?

torfjelde:
Hmm, it's unclear to me whether it's worth adding these methods when they have "no use" unless some notion of conditioning is also added 😕 

> How is https://github.com/TuringLang/AbstractMCMC.jl/pull/86 related to this PR?

`getparams` is probably overlapping between the two PRs, but the `recompute_logprob!!` method is not

src/AbstractMCMC.jl
```diff
@@ -80,6 +80,54 @@ The `MCMCSerial` algorithm allows users to sample serially, with no thread or pr
...
+"""
+function decondition end
+
+"""
+    condition(model, observations)
+
+Condition the generative model `model` on some observed data, creating a new model of the (possibly
+unnormalized) posterior distribution over them.
+
+`observations` can be of any supported internal trace type, or a fixed probability expression.
```
torfjelde, line 105: There's no notion of "trace type" in AbstractMCMC

src/AbstractMCMC.jl
```diff
@@ -80,6 +80,54 @@ The `MCMCSerial` algorithm allows users to sample serially, with no thread or pr
...
+Recompute the log-probability of the `model` based on the given `state` and return the resulting state.
+"""
+function recompute_logprob!!(rng, model, sampler, state) end
+
+"""
+    getparams(state)
+
+Returns the values of the parameters in the state.
+"""
+function getparams(state) end
```
torfjelde, line 129: Do we say anything about what the form of these parameters should be?

torfjelde:
I'm still a bit uncertain about all of this tbh. I feel like right now we're just shoving `condition` and `decondition` (which I don't think we need for Gibbs?) into AbstractMCMC.jl to motivate the inclusion of `recompute_logprob!!` 😅 

I think if this is the case, then I'm preferential to ignoring my original comment of "needing `condition` to motivate `recompute_logprob!!`", i.e. just leave it as you did originally (without `condition` and `decondition`).

torfjelde:
> I also think we should add something like AbstractState to normalize the design of state. This will introduce types for state everywhere, I am unsure of the impact. What's your thoughts on this?

Not for this PR at least:) If we want to discuss this, then we should open an issue and move discussion there.

torfjelde:
> The latter will technically change the interface, so I bumped the minor version.

It seems you've bumped the _major_ version, not the minor version?

Also, if we're making this part of the interface, we should probably document this?

docs/src/gibbs.md
```diff
@@ -0,0 +1,404 @@
...
+    end
+    return GibbsTransition(initial_params), GibbsState(initial_params, states)
+end
+
+function AbstractMCMC.step(
+    rng::AbstractRNG,
+    logdensity_model::AbstractMCMC.LogDensityModel,
+    spl::Gibbs,
+    state::GibbsState,
+    args...;
```
torfjelde, line 313:
```suggestion
```
IIRC `args...` should not be there?

docs/src/gibbs.md
```diff
@@ -0,0 +1,404 @@
...
+    end
+end
+```
+
+At last, we can proceed to implement the Gibbs sampler.
+
+```julia
+struct Gibbs <: AbstractMCMC.AbstractSampler
+    sampler_map::OrderedDict
+end
```
torfjelde, line 263:
```suggestion
struct Gibbs{D} <: AbstractMCMC.AbstractSampler
    sampler_map::D
end
```

docs/src/gibbs.md
```diff
@@ -0,0 +1,404 @@
...
+
+struct GibbsState
+    vi::NamedTuple
+    states::OrderedDict
+end
+
+struct GibbsTransition
+    values::NamedTuple
+end
```
torfjelde, line 272:
```suggestion
struct GibbsTransition{NT<:NamedTuple}
    values::NT
end
```

docs/src/gibbs.md
```diff
@@ -0,0 +1,404 @@
...
+
+At last, we can proceed to implement the Gibbs sampler.
+
+```julia
+struct Gibbs <: AbstractMCMC.AbstractSampler
+    sampler_map::OrderedDict
+end
+
+struct GibbsState
+    vi::NamedTuple
```
torfjelde, line 266: I saw the comment about calling this `trace` instead, but why not just call this `realization` or `values`?

docs/src/gibbs.md
```diff
@@ -0,0 +1,404 @@
...
+Next we define the four `state` interface functions.
+
+```julia
+AbstractMCMC.get_params(state::MHState) = state.params
+AbstractMCMC.set_params!!(state::MHState, params) = MHState(params, state.logp)
+AbstractMCMC.get_logprob(state::MHState) = state.logp
+AbstractMCMC.set_logprob!!(state::MHState, logp) = MHState(state.params, logp)
+```
+
+These are the functions that was used in the `recompute_logprob!!` function above.
```
torfjelde, line 160: This isn't quite true, no? Only two of them are actually used.

docs/src/gibbs.md
```diff
@@ -0,0 +1,404 @@
...
+function LogDensityProblems.capabilities(::ConditionedHierNormal)
+    return LogDensityProblems.LogDensityOrder{0}()
+end
+```
+
+the model should also define a function that allows the recomputation of the log probability given a sampler state.
+The reason for this is that, when we break down the joint probability into conditional probabilities, individual conditional probability problems are conditional on the values of the other variables.
+Between the Gibbs sampler sweeps, the values of the variables may change, and we need to recompute the log probability of the current state.
+
+A recomputation function could use the `state` interface to return a new state with the updated log probability.
```
torfjelde, line 108: Might be useful to add a comment on why this is useful, e.g. "later we will see how this functionality is useful if we want to implemented higher-order samplers such as Gibbs sampling"

src/gibbs.jl
```diff
@@ -0,0 +1,225 @@
...
+"""
+struct Gibbs <: AbstractMCMC.AbstractSampler
+    sampler_map::NamedTuple
+    parameter_names::Tuple{Vararg{Symbol}}
+
+    function Gibbs(sampler_map::NamedTuple)
+        parameter_names = Tuple(keys(sampler_map))
+        return new(sampler_map, parameter_names)
+    end
+end
```
torfjelde, line 17:
```suggestion
struct Gibbs{NT} <: AbstractMCMC.AbstractSampler
    sampler_map::NT
end
```

`parameter_names` seems wholly unnecessary here?

src/gibbs.jl
```diff
@@ -0,0 +1,225 @@
...
+
+struct GibbsState
```
torfjelde, line 19: What's the reason for not making this typed?

src/gibbs.jl
```diff
@@ -0,0 +1,225 @@
...
+    """
+    mcmc_states::NamedTuple
+
+    """
+    `variable_sizes` maps parameters to their sizes.
+    """
+    variable_sizes::NamedTuple
+end
+
+struct GibbsTransition
```
torfjelde, line 36: Same here

src/gibbs.jl
```diff
@@ -0,0 +1,225 @@
...
+        sub_state = gibbs_state.mcmc_states[parameter_variable]
+        trace = merge(
+            trace,
+            unflatten(
+                vec(sub_state),
+                NamedTuple{(parameter_variable,)}((
+                    gibbs_state.variable_sizes[parameter_variable],
+                )),
+            ),
+        )
```
torfjelde, line 112: Would generally recommend against this sort of "nesting" functions in arguments; it makes things less readable than if you just added a few lines above constructing the arguments.

src/gibbs.jl
```diff
@@ -0,0 +1,225 @@
...
+    end
+    return trace
+end
+
+function AbstractMCMC.step(
+    rng::Random.AbstractRNG,
+    logdensity_model::AbstractMCMC.LogDensityModel,
+    sampler::Gibbs,
+    args...;
+    initial_params::NamedTuple,
```
torfjelde, line 122: This does break with what we're generally suggesting in other AbstractMCMC.jl places, no? AFAIK we're generally assuming / suggesting that `initial_params` should be `AbstractVector`

torfjelde: Maybe worth adding a handler which can do either `NamedTuple` or `AbstractVector`

src/gibbs.jl
```diff
@@ -0,0 +1,225 @@
...
+            Tuple([initial_params[g] for g in variables_to_be_conditioned_on])
+        )
+        sub_problem_parameters_values = NamedTuple{(parameter_variable,)}((
+            initial_params[parameter_variable],
+        ))
+
+        # LogDensityProblems' `logdensity` function expects a single vector of real numbers
+        # `Gibbs` stores the parameters as a named tuple, thus we need to flatten the sub_problem_parameters_values
+        # and unflatten after the sampling step
+        variable_sizes[parameter_variable] = Tuple(size(initial_params[parameter_variable]))
```
torfjelde, line 151:
```suggestion
        variable_sizes[parameter_variable] = size(initial_params[parameter_variable])
```
`Tuple` should really not be necessary here?

src/gibbs.jl
```diff
@@ -0,0 +1,225 @@
...
+                    AbstractMCMC.condition(
+                        logdensity_model.logdensity, conditioning_variables_values
+                    ),
+                ),
+                sub_sampler,
+                args...;
+                initial_params=flattened_sub_problem_parameters_values,
+                kwargs...,
+            ),
+        )
```
torfjelde, line 167: As as above, much prefer the following

```suggestion
        sub_model = AbstractMCMC.LogDensityModel(
            AbstractMCMC.condition(
                logdensity_model.logdensity, conditioning_variables_values
            ),
        )
        sub_state = last(
            AbstractMCMC.step(
                rng,
                sub_model,
                sub_sampler,
                args...;
                initial_params=flattened_sub_problem_parameters_values,
                kwargs...,
            ),
        )
```

src/gibbs.jl
```diff
@@ -0,0 +1,225 @@
...
+    sampler::Gibbs,
+    gibbs_state::GibbsState,
+    args...;
+    kwargs...,
+)
+    trace = gibbs_state.trace
+    mcmc_states = gibbs_state.mcmc_states
+    variable_sizes = gibbs_state.variable_sizes
+
+    mcmc_states_dict = Dict(
```
torfjelde, line 190: Is the usage of `Dict` here intentional?

torfjelde:
> I'd encourage @torfjelde and @sunxd3 to have more discussions to converge on a sensible Gibbs design; we will never get a good implementation if we don't start working on it.

Definitively! But I think it's also partially a thing that we need to try out a bit, and we're not going to get it right on the first try, hence my reluctance to putting an impl in AbstractMCMC.jl as soon as we enable the functionality.

sunxd3:
the test error seems to be Julia 1.6-only related

sunxd3:
related reply from @devmotion https://github.com/TuringLang/Turing.jl/pull/2304#issuecomment-2291097709

torfjelde (changes requested):
I'm very much in favour of using explicit functions for the interface and not overloading methods (in ways that the original methods were not intended for) :confused:

torfjelde:
I do think the entire process of this would be quite a bit less painful if we did the following (I believe I've mentioned this before; if not, I apologize):
1. Improve #86 to a finalized form . This is useful, not just for Gibbs sampling.
2. Make a separate package, e.g. `AbstractMCMCGibbs.jl`, which implements the Gibbs-only stuff, e.g. `recompute_logprob!!` and the sampler mapping stuff.

This is how we're doing it with MCMCTempering.jl, i.e. keep it as a separate package and slowly move pieces to AbstractMCMC.jl if it seems suitable. My problem, as stated before, is that the current Gibbs impls we're working with are really not good enough as I think is evident by a) issues that we've encountered with my Turing.jl-impl in https://github.com/TuringLang/Turing.jl/pull/2328#issuecomment-2378971734, and b) the amount of iterating you've done in this PR. This shit is complicated :grimacing: And I imagine it's really annoying iterating on this back and forth but without actually getting stuff merged.. 

So, I think a separate package would just make this entire process much easier @sunxd3 ; then we can iterate much faster on ideas (just make breaking releases), and then we can just upstream changes as we finalize things there.

design_notes/on_gibbs_implementation.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Proposed Interface
+
+The two functions `getparams` and `recompute_logp!!` form a minimal interface to support the `Gibbs` implementation. However, there are concerns about introducing them directly into `AbstractMCMC`. The main reason is that `AbstractMCMC` is a root dependency of the `Turing` packages, so we want to be very careful with new releases.
+
+Here, some alternative functions that achieve the same functionality as `getparams` and `recompute_logp!!` are proposed, but without introducing new interface functions.
+
+For `getparams`, we can use `Base.vec`. It is a `Base` function, so there's no need to export anything from `AbstractMCMC`. Since `getparams` should return a vector, using `vec` makes sense. The concern is that, officially, `Base.vec` is defined for `AbstractArray`, so it remains a question whether we should only introduce `vec` in the absence of other `AbstractArray` interfaces.
```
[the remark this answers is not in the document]
sunxd3: I can resonate, issue with `public` is they still count as public interface, unsure if we need to make minor release

design_notes/on_gibbs_implementation.md
```diff
@@ -0,0 +1,59 @@
...
+
+For `recompute_logp!!`, we could overload `LogDensityProblems.logdensity(logdensity_model::AbstractMCMC.LogDensityModel, state::State; recompute_logp=true)` to compute the log probability. If `recompute_logp` is `true`, it should recompute the log probability of the state. Otherwise, it could use the log probability stored in the state. To allow updating the log probability stored in the state, samplers should define outer constructor for their state types `StateType(state::StateType, logdensity=logp)` that takes an existing `state` and a log probability value `logp`, and returns a new state of the same type with the updated log probability.
+
+While overloading `LogDensityProblems.logdensity` to take a state object instead of a vector for the second argument somewhat deviates from the interface in `LogDensityProblems`, it provides a clean and extensible solution for handling log probability recomputation within the existing interface.
```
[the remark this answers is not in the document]
sunxd3: mainly to not make any changes to the interface, so these are just "recommendations"

I think `AbstractMCMC` depends on `LogDensityProblems`

torfjelde:
https://github.com/TuringLang/AbstractMCMC.jl/issues/85#issuecomment-2061300622

PR #146, opened 2024-09-11
Add error hint for MethodErrors on LogDensityModel

penelopeysm:
If a user creates their own AbstractMCMC sampler and passes it something that obeys the LogDensityProblems.jl interface, AbstractMCMC quietly wraps it in a `LogDensityModel` type which doesn't obey the same interface, which can be mildly confusing. (Well, it did catch me out, so there may be other new users out there too who bumped into it too :))

As this is a conscious decision (cf. https://github.com/TuringLang/AbstractMCMC.jl/pull/110#discussion_r1043893301) this PR makes it such that if somebody tries to call LogDensityProblems.jl methods on a LogDensityModel, they get a friendly nudge in the right direction.

MWE:

```julia
julia> using LogDensityProblems

julia> using AbstractMCMC: LogDensityModel

julia> struct MyModel end

julia> # In practice one would define more methods, but this is the only one needed for AbstractMCMC to not complain

julia> LogDensityProblems.capabilities(::Type{MyModel}) = LogDensityProblems.LogDensityOrder{0}()

julia> lm = LogDensityModel(MyModel())
LogDensityModel{MyModel}(MyModel())

julia> LogDensityProblems.logdensity(lm, 0)
ERROR: MethodError: no method matching logdensity(::LogDensityModel{MyModel}, ::Int64)
AbstractMCMC.LogDensityModel is a wrapper and does not itself implement the LogDensityProblems.jl interface. To use LogDensityProblems.jl methods, access the inner type with (e.g.) `logdensity(model.logdensity, params)` instead of `logdensity(model, params)`.
Stacktrace:
 [1] top-level scope
   @ REPL[7]:1
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,9 +1,9 @@
 name = "AbstractMCMC"
 uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
-keywords = ["markov chain monte carlo", "probablistic programming"]
+keywords = ["markov chain monte carlo", "probabilistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "5.2.0"
+version = "5.2.1"
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probabilistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "5.2.1"
+version = "5.3.0"
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
--- a/src/AbstractMCMC.jl
+++ b/src/AbstractMCMC.jl
@@ -88,4 +88,19 @@ include("stepper.jl")
 include("transducer.jl")
 include("logdensityproblems.jl")
 
+function __init__()
+    if isdefined(Base.Experimental, :register_error_hint)
+        Base.Experimental.register_error_hint(MethodError) do io, exc, argtypes, _
+            if Base.parentmodule(exc.f) == LogDensityProblems &&
+                any(a -> a <: LogDensityModel, argtypes)
+                printstyled(
+                    io,
+                    "\nAbstractMCMC.LogDensityModel is a wrapper and does not itself implement the LogDensityProblems.jl interface. To use LogDensityProblems.jl methods, access the inner type with (e.g.) `logdensity(model.logdensity, params)` instead of `logdensity(model, params)`.";
+                    color=:cyan,
+                )
+            end
+        end
+    end
+end
+
 end # module AbstractMCMC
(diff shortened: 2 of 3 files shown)


src/AbstractMCMC.jl
```diff
@@ -88,4 +88,19 @@ include("stepper.jl")
 include("transducer.jl")
 include("logdensityproblems.jl")
 
+function __init__()
+    if isdefined(Base.Experimental, :register_error_hint)
```
devmotion, line 92: A tiny performance improvement but IIRC `__init__()` can contribute to pre-compilation and loading times to a surprising extent, so in older Julia versions without error hints you can skip it completely by swapping these two lines:
 ```suggestion
function __init__()
    if isdefined(Base.Experimental, :register_error_hint)
```
```suggestion
if isdefined(Base.Experimental, :register_error_hint)
    function __init__()
```

src/AbstractMCMC.jl
```diff
@@ -88,4 +88,19 @@ include("stepper.jl")
...
+        Base.Experimental.register_error_hint(MethodError) do io, exc, argtypes, _
+            if Base.parentmodule(exc.f) == LogDensityProblems &&
+                any(a -> a <: LogDensityModel, argtypes)
+                printstyled(
+                    io,
+                    "\nAbstractMCMC.LogDensityModel is a wrapper and does not itself implement the LogDensityProblems.jl interface. To use LogDensityProblems.jl methods, access the inner type with (e.g.) `logdensity(model.logdensity, params)` instead of `logdensity(model, params)`.";
+                    color=:cyan,
```
devmotion, line 99: I think it's safer to just print a vanilla message, without setting any color - you never know what colors people use in their REPL, so some might not be able to read text if it is displayed in an unfortunate/mismatching color.

```suggestion
                print(
                    io,
                    "\n`AbstractMCMC.LogDensityModel` is a wrapper and does not itself implement the LogDensityProblems.jl interface. To use LogDensityProblems.jl methods, access the inner type with (e.g.) `logdensity(model.logdensity, params)` instead of `logdensity(model, params)`.",
```

penelopeysm:
(Thanks on both points :))

Project.toml
```diff
@@ -1,9 +1,9 @@
 name = "AbstractMCMC"
 uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
-keywords = ["markov chain monte carlo", "probablistic programming"]
+keywords = ["markov chain monte carlo", "probabilistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "5.2.0"
+version = "5.2.1"
```
devmotion, line 6: I think it's not a bug fix but rather a feature, so I'd make it
```suggestion
version = "5.3.0"
```

devmotion (commented):
Maybe it would be good to add a (very simple) test as well to ensure it works as intended?

Project.toml
penelopeysm: Ooh, okay, I thought of it more as a bug fix ('bug' being that error message is unclear). But thinking about it a bit more this does modify the package's API in that people can catch the error and check the message, so will bump the minor.

devmotion (approved):

PR #147, opened 2024-10-04
Use `copy` instead of `deepcopy` for RNGs

devmotion:
My take-away from https://github.com/JuliaLang/julia/issues/42899 is that we should use `copy` instead of `deepcopy` for RNGs. More generally, based on https://github.com/JuliaLang/julia/issues/42899#issuecomment-1058487002 we might also want to use `copy` for models and samplers (possibly requires defining `copy` in a few downstream packages).


sunxd3:
interesting, thanks David

sunxd3 (approved):

PR #148, opened 2024-10-15
Add a note on future interface for `Gibbs`

sunxd3:
This is a summary for the closed #144. 

The note should provide some contextual information. Suggestions and discussions are very welcomed.

### DIFF
--- a/design_notes/interface_for_gibbs.md
+++ b/design_notes/interface_for_gibbs.md
@@ -1,0 +1,77 @@
+# Notes on Potential Future Interface for Gibbs Sampling Support
+
+## Background
+
+This document was written after [PR #144](https://github.com/TuringLang/AbstractMCMC.jl/pull/144) was closed.
+
+It was last updated on October 15, 2024. At that time:
+
+- _AbstractMCMC.jl_ was on version 5.5.0  
+- _Turing.jl_ was on version 0.34.1
+
+The goal is to document some of the considerations that went into the closed PR mentioned above.
+
+## Gibbs Sampling Considerations
+
+### Recomputing Log Densities for Parameter Groups
+
+Let's consider splitting the model parameters into three groups (assuming the grouping stays fixed between iterations). Each parameter group will have a corresponding sampler state (along with the sampler used for that group).
+
+In the general case, the log densities stored in the states will be incorrect at the time of sampling each group. This is because the values of the other two parameter groups can change from when the current log density was computed, as they get updated within the Gibbs sweep.
+
+### Current Approach: `recompute_logp!!`
+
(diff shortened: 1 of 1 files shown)


design_notes/interface_for_gibbs.md
```diff
@@ -0,0 +1,77 @@
...
+- _AbstractMCMC.jl_ was on version 5.5.0  
+- _Turing.jl_ was on version 0.34.1
+
+The goal is to document some of the considerations that went into the closed PR mentioned above.
+
+## Gibbs Sampling Considerations
+
+### Recomputing Log Densities for Parameter Groups
+
+Let's consider splitting the model parameters into three groups (assuming the grouping stays fixed between iterations). Each parameter group will have a corresponding sampler state (along with the sampler used for that group).
```
sunxd3, line 18:
```suggestion
Let's consider splitting the model parameters into several groups (assuming the grouping stays fixed between iterations). Each parameter group will have a corresponding sampler state (along with the sampler used for that group).
```

design_notes/interface_for_gibbs.md
```diff
@@ -0,0 +1,77 @@
...
+
+To complement this computation interface, we would need functions to `get` and `set` the log likelihood and log prior from/to the sampler states.
+
+For situations where model-specific details are required to compute the log density from a sampler state, the necessary abstractions are not yet clear. We will need to consider appropriate abstractions as such use cases emerge.
+
+## Additional Notes on a More Independent Gibbs Implementation
+
+### Regarding `AbstractPPL.condition`
+
+While the `condition` function is a promising idea for Gibbs sampling, it is not currently being utilized in _Turing.jl_'s implementation. Instead, _Turing.jl_ uses a `GibbsContext` for reasons outlined [here](https://github.com/TuringLang/Turing.jl/blob/3c91eec43176d26048b810aae0f6f2fac0686cfa/src/experimental/gibbs.jl#L1-L12). Additionally, _JuliaBUGS_ requires caching the Markov blanket when calling `condition`, which means the proposed `Gibbs` implementation in this PR would not be fully compatible.
```
sunxd3, line 69:
```suggestion
While the `condition` function is a promising idea for Gibbs sampling, it is not currently being utilized in _Turing.jl_'s implementation. Instead, _Turing.jl_ uses a `GibbsContext` for reasons outlined [here](https://github.com/TuringLang/Turing.jl/blob/3c91eec43176d26048b810aae0f6f2fac0686cfa/src/experimental/gibbs.jl#L1-L12). Additionally, _JuliaBUGS_ requires caching the Markov blanket when calling `condition`, which means the proposed `Gibbs` implementation in the PR above would not be fully compatible.
```

sunxd3:
recent PRs like https://github.com/TuringLang/AbstractMCMC.jl/pull/150 and discussions like https://github.com/TuringLang/AbstractPPL.jl/discussions/108 supersede this, so close now

PR #150, opened 2024-10-23
Add optional argument of `logdensity_function` to `getparams` and `setparams!!`

sunxd3:
ref: https://github.com/TuringLang/AdvancedHMC.jl/pull/378#discussion_r1811326624

### DIFF
--- a/src/AbstractMCMC.jl
+++ b/src/AbstractMCMC.jl
@@ -81,23 +81,27 @@ The `MCMCSerial` algorithm allows users to sample serially, with no thread or pr
 struct MCMCSerial <: AbstractMCMCEnsemble end
 
 """
-    getparams(state[; kwargs...])
+    getparams(state[, logdensity_function])
 
 Retrieve the values of parameters from the sampler's `state` as a `Vector{<:Real}`.
 """
 function getparams end
 
 """
-    setparams!!(state, params)
+    setparams!!(state, params[, logdensity_function])
 
 Set the values of parameters in the sampler's `state` from a `Vector{<:Real}`. 
 
 This function should follow the `BangBang` interface: mutate `state` in-place if possible and 
 return the mutated `state`. Otherwise, it should return a new `state` containing the updated parameters.
 
-Although not enforced, it should hold that `setparams!!(state, getparams(state)) == state`. In another
-word, the sampler should implement a consistent transformation between its internal representation
+Although not enforced, it should hold that `setparams!!(state, getparams(state)) == state`. In other
+words, the sampler should implement a consistent transformation between its internal representation
 and the vector representation of the parameter values.
+
+Sometimes, to maintain the consistency of the log density and parameter values, an optional
+`logdensity_function` can be provided. This is useful for samplers that need to evaluate the
+log density at the new parameter values.
 """
 function setparams!! end
 


sunxd3:
I kept `logdensity_function` as the last argument so that it can be optional

torfjelde:
> I left some comments above. In addition, we want to consider using model::AbstractModel consistently for these APIs, since we seem to be doing them already for the sample function. See e.g., [here](https://github.com/TuringLang/AbstractMCMC.jl/blob/467b07621975d81856792d0788f0e490c3ecaaa5/src/logdensityproblems.jl#L110).

Agree with this. This is what I've done in MCMCTempering.jl.

torfjelde (changes requested):
I second @yebai 's suggestions on the type of model passed; it should be `AbstractMCMC.AbstractModel`, not `logdensity_function`. A `logdensity_function` is indicated by `AbstracftMCMC.LogDensityModel`.

Moreover, IMO we should pass `model` as the first argument, not the last, and then "drop it" by default, i.e.
```julia
setparams!!(model, state, params) = setparams!!(state, params)
```
That way, if one needs access to the `model` you overload `setparams(model, state, params)` directly for that specific model type; otherwise, you just overload `setparams!!(state, params)`. This will make it less likely that we'll run into method ambiguities, since the `setparams!!(model, state, params)` will be specific for a given model type.

sunxd3:
Updated @yebai @torfjelde

sunxd3:
At the moment, I think pass them as two argument is more preferable for me -- just a less layer of indirections.

sunxd3:
Thanks Hong, for AbstractMCMC, let's try to have two approvals, so I'll wait a bit for @torfjelde's opinion

Project.toml
```diff
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probabilistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "5.5.0"
+version = "5.5.1"
```
sunxd3, line 6:
```suggestion
version = "5.6.0"
```
maybe?

penelopeysm: Yeah, I guess so! Even if the new functionality introduced is very trivial.

src/AbstractMCMC.jl
```diff
@@ -81,26 +81,51 @@ The `MCMCSerial` algorithm allows users to sample serially, with no thread or pr
...
+    getparams(state)
 
 Retrieve the values of parameters from the sampler's `state` as a `Vector{<:Real}`.
 """
 function getparams end
 
+function getparams(logdensity, state)
+    return getparams(_model(logdensity), state)
+end
+
```
torfjelde, line 95:
```suggestion
```
Is there a reason why we want this?
From what I can telll, we don't. The way we do it in AbstractMCMC.jl is that a ` LogDensityModel` indicates that we're working with a `logdensity_function`, hence if you want to overload this for, say, `DynamicPPL.LogDensityFunction` you can just overload `getparams(model::LogDensityModel{<:DynamicPPL.LogDensityFunction}, ...)`; no need to add another method instance.

src/AbstractMCMC.jl
```diff
@@ -81,26 +81,51 @@ The `MCMCSerial` algorithm allows users to sample serially, with no thread or pr
 struct MCMCSerial <: AbstractMCMCEnsemble end
 
 """
-    getparams(state[; kwargs...])
+    getparams(model::AbstractModel, state)
+    getparams(logdensity, state)
+    getparams(state)
```
torfjelde, line 86:
```suggestion
    getparams([model::AbstractModel, ]state)
```

torfjelde: See comment below for removal of `logdenisty` method .

src/AbstractMCMC.jl
```diff
@@ -81,26 +81,51 @@ The `MCMCSerial` algorithm allows users to sample serially, with no thread or pr
...
+end
+
+function getparams(model::AbstractModel, state)
+    return getparams(state)
+end
+
 """
+    setparams!!(model::AbstractModel, state, params)
+    setparams!!(logdensity, state, params)
     setparams!!(state, params)
```
torfjelde, line 103:
```suggestion
    setparams!!([model::AbstractModel, ]state, params)
```

src/AbstractMCMC.jl
```diff
@@ -81,26 +81,51 @@ The `MCMCSerial` algorithm allows users to sample serially, with no thread or pr
...
 This function should follow the `BangBang` interface: mutate `state` in-place if possible and 
 return the mutated `state`. Otherwise, it should return a new `state` containing the updated parameters.
 
-Although not enforced, it should hold that `setparams!!(state, getparams(state)) == state`. In another
-word, the sampler should implement a consistent transformation between its internal representation
+Although not enforced, it should hold that `setparams!!(state, getparams(state)) == state`. In other
+words, the sampler should implement a consistent transformation between its internal representation
 and the vector representation of the parameter values.
+
+Sometimes, to maintain the consistency of the log density and parameter values, a `model::AbstractModel`
```
torfjelde, line 114:
```suggestion
Sometimes, to maintain the consistency of the log density and parameter values, a `model`
```

Type of `model` is already specified

src/AbstractMCMC.jl
```diff
@@ -81,26 +81,51 @@ The `MCMCSerial` algorithm allows users to sample serially, with no thread or pr
...
+should be provided. This is useful for samplers that need to evaluate the log density at the new parameter values.
+If `model` is not an `AbstractMCMC.AbstractModel`, by default, it is assumed to be a log density function following 
+the `LogDensityProblems.jl` interface, and will be wrapped with [`AbstractMCMC.LogDensityModel`](@ref).
```
torfjelde, line 117:
```suggestion
```

torfjelde: This should be done by the user. Remember, `LogDensityModel` should be used for all AbstractMCMC.jl methods wher a logdensity function is the underlying.

src/AbstractMCMC.jl
```diff
@@ -81,26 +81,51 @@ The `MCMCSerial` algorithm allows users to sample serially, with no thread or pr
...
 """
 function setparams!! end
 
+function setparams!!(logdensity, state, params)
+    return setparams!!(_model(logdensity), state, params)
+end
+
```
torfjelde, line 124:
```suggestion
```

torfjelde (changes requested):
Looking good! But IMO we should not have additional implementations for a `logdensity_function` arguments. Currently, we take the approach of passing a `LogDensityModel` everywhere we want to indicate that we're working with a logdensity function; we should follow this convention in `setparams!!` and `getparams` too.

sunxd3:
No objection removing them, I added them because this what `sample` does. But no real reason for adding something we don't need now.

torfjelde:
> No objection removing them, I added them because this what sample does. But no real reason for adding something we don't need now.

Makes sense:) So `sample` is meant to be end-user facing, and so we want to make that as easy as possible. However, the methods meant for inference devs, e.g. `step` are all assuming a logdensity function is already wrapped in a `LogDensityModel`. If you're using `setparams!!` or `getparams`, this means you're already accessing `step` and so I'm thinking we should follow the same convention as `step` 👍

sunxd3:
very much agree 👍

torfjelde (approved):
I'll approve because it's in a good enough state now 👍 
But is how to implement this documented somewhere? And if so, should we add a note to make it clear that if you don't need access to the `model` you should implement `setparams!!(state, params)` and not `setparams!!(model, state, params)` ?

sunxd3:
I assume [here](https://github.com/TuringLang/MCMCTempering.jl/blob/deb96684496f3fbd011b9f70f28c49a161def23f/ext/MCMCTemperingAdvancedHMCExt.jl#L12 ) `model` follows `LogDensityProblems`?

torfjelde:
No, it's a `AbstractMCMC.LogDensityModel`

sunxd3:
> But is how to implement this documented somewhere? And if so, should we add a note to make it clear that if you don't need access to the model you should implement setparams!!(state, params) and not setparams!!(model, state, params) ?

We definitely should, but I am not very clear on what directions to give now. I would vote for just merge this and enhance this part of the documentation later (also the interface might change again in near future). Do you think this is a bad idea?

torfjelde:
> We definitely should, but I am not very clear on what directions to give now. I would vote for just merge this and enhance this part of the documentation later (also the interface might change again in near future). Do you think this is a bad idea?

I'm happy with this given that right now my reviews are quite async (this will be better starting on Thu this coming week) since I'm in a different timezone. 

Regarding the directions, I stated above

>  And if so, should we add a note to make it clear that if you don't need access to the model you should implement setparams!!(state, params) and not setparams!!(model, state, params) ?

That's the main thing IMO 🤷

torfjelde:
Good stuff @sunxd3 ! :)

PR #155, opened 2025-01-08
Update CI

penelopeysm:
- Bumping versions of actions

- Also changing 1.6 -> 'min' so that we don't have to update that in the future

- Also changing the 'nightly' test (brittle) to a new prerelease workflow

- Also enabling concurrency on PRs so that outdated CI runs are cancelled

This is all in line with other repos in TuringLang

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -18,36 +18,24 @@ concurrency:
 
 jobs:
   test:
-    runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.version == 'nightly' }}
+    runs-on: ${{ matrix.runner.os }}
     strategy:
       matrix:
         version:
           - 'min'
           - '1'
-        os:
-          - ubuntu-latest
-          - macOS-latest
-          - windows-latest
-        arch:
-          - x86
-          - x64
         num_threads:
           - 1
           - 2
-        exclude:
-          - os: windows-latest
-            arch: x86
-          - os: macOS-latest
-            arch: x86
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -39,25 +48,31 @@ jobs:
...
+
     steps:
-      - uses: actions/checkout@v2
-      - uses: julia-actions/setup-julia@v1
+      - uses: actions/checkout@v4
+
+      - uses: julia-actions/setup-julia@v2
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
```
devmotion, line 58: This should be updated as well since currently we test `x64` on Apple Silicon runners. There are multiple options here, the main question I guess is whether we want to keep testing Julia 32bit. If not, the easiest fix is just to remove `arch` and `matrix.arch` completely. Alternatively, we could use a default of `matrix.arch: ''` and change this line to
```suggestion
          arch: ${{ (matrix.arch == '') && runner.arch || matrix.arch }}
```
which means that by default the Julia version that matches the runner's architecture is used (and hence aarch64 on Apple Silicon) but if x86 is requested explicitly it would still be used. See e.g. https://discourse.julialang.org/t/how-to-fix-github-actions-ci-failures-with-julia-1-6-or-1-7-on-macos-latest-and-macos-14/117019

penelopeysm: Ooh good spot! we changed that on Turing and DPPL but I forgot here.

PR #158, opened 2025-02-10
Add `getadtype` function to AbstractSampler interface

penelopeysm:
Right now, Turing.jl contains some pretty ad hoc code to determine the underlying AD backend for a sampler:

https://github.com/TuringLang/Turing.jl/blob/ddd74b1cf4694b66698f225db549d2cfb5eb826c/src/mcmc/Inference.jl#L181-L189

In the process, it does a bunch of type piracy and this definition also makes it hard to move any of this behaviour to DynamicPPL.

In particular, it actually means that the AD backend (inside a `LogDensityProblemsAD.ADgradient` object) is only specified by code inside Turing.jl, even though DynamicPPL should have enough information to infer the AD type from the sampler it's given.

This PR proposes to move the AD type specification to the lower layer of AbstractMCMC so that this information is accessible across more layers.

It's not 100% obvious to me that gradient-based samplers _necessarily_ need to declare their adtype (maybe depending on the sampler implementation, it might be declared elsewhere, e.g. in the model), but having `getadtype` be a property of the sampler does at least match what we currently do.

**Followups if this is merged**

- [x] Implement `getadtype(::DynamicPPL.Sampler)`
- [ ] Remove extra code from Turing
- [ ] (lower priority) Implement in AdvancedHMC

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probabilistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "5.6.1"
+version = "5.7.0"
 
 [deps]
 BangBang = "198e06fe-97b7-11e9-32a5-e1d131e6ad66"
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -20,14 +20,18 @@ chainsstack(c) = c
 chainsstack(c::AbstractVector{<:AbstractChains}) = reduce(chainscat, c)
 
 """
-    getadtype(sampler::AbstractSampler)
+    getadtype(s::AbstractSampler)
+    getadtype(m::AbstractModel, s::AbstractSampler)
 
-If the sampler specifies an automatic differentiation (AD) backend to use, this
-function should return the corresponding `ADTypes.AbstractADType`.
+Specify the `ADTypes.AbstractADType` to be used when sampling from model `m` using sampler `s`.
+
+If the model is not relevant, then the implementation of AbstractSampler can
+directly overload the single-argument method `getadtype(s::AbstractSampler)`.
 
 By default, this returns `nothing`.
 """
 getadtype(::AbstractSampler) = nothing
+getadtype(::AbstractModel, spl::AbstractSampler) = getadtype(spl)
 
 """
     bundle_samples(samples, model, sampler, state, chain_type[; kwargs...])
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -20,6 +20,16 @@ chainsstack(c) = c
 chainsstack(c::AbstractVector{<:AbstractChains}) = reduce(chainscat, c)
 
 """
+    getadtype(sampler::AbstractSampler)
+
+If the sampler specifies an automatic differentiation (AD) backend to use, this
+function should return the corresponding `ADTypes.AbstractADType`.
+
+By default, this returns `nothing`.
+"""
+getadtype(::AbstractSampler) = nothing
+
+"""
     bundle_samples(samples, model, sampler, state, chain_type[; kwargs...])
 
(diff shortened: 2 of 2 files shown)


penelopeysm:
> So, I'd suggest we use getadtype(LogDensityModel, Sampler) instead of getadtype(Sampler) for the interface

I support that:) One thing, we don't actually use `LogDensityModel` in Turing, so I think we want to keep it general with

```julia
getadtype(::AbstractModel, ::AbstractSampler)
```

And I wonder if we could also maybe use the same strategy as getparams and have a default implementation

```julia
getadtype(::AbstractModel, spl::AbstractSampler) = getadtype(spl)
```

and let users overload the method they prefer to use?

mhauru:
I'd be happy with the
```julia
getadtype(::AbstractModel, spl::AbstractSampler) = getadtype(spl)
```
approach.

What's the convention for dependencies of interface packages? I feel a bit awkward about explicitly requiring that the return type is `ADTypes.AbstactADType`, but not having `ADTypes` as a dependency. If Julia would allow us to enforce the interface, we definitely would have it as a dependency. But can also see this both ways, not wanting to add a dependency that the code doesn't use (only the docs do).

penelopeysm:
The more I think about it, the more I'm convinced that the adtype should be in the model, not the sampler. But at the same time I don't know if it would be worth it to break everything :/

(1)

From a gradient-based sampler's point of view, it should be agnostic to how the gradient is calculated, it just cares that it can have a gradient. For example, if the model obeys the LogDensityProblems interface, the sampler only really cares that `logdensity_and_gradient(model, params)` works; it shouldn't care about the implementation of this.

Indeed, one might find that for a particular model it is possible to implement `logdensity_and_gradient` analytically, in which case we can still use HMC just fine even without any AD backend. (The AdvancedHMC-specific `sample` interface [allows for this](https://github.com/TuringLang/AdvancedHMC.jl?tab=readme-ov-file#gradients), but not `Turing.HMC`.)

(2)

The best AD backend to use is almost entirely determined by the model (e.g. the number of parameters, or the functionality inside the model) and thus it makes sense that the choice of AD backend should be tied to the model.

(3)

In DynamicPPL we might often want to test AD correctness and/or performance with a model, in a way like this

```julia
@model f() = ...

ref_logp, ref_grad = logdensity_and_gradient_with(f(), AutoForwardDiff())
logp, grad = logdensity_and_gradient_with(f(), AutoMooncake())\
@test isapprox(logp, ref_logp)
@test isapprox(grad, ref_grad)
```

We definitely don't want to have to construct a dummy sampler just to pass the adtype to the model.

(4)

Currently the way that Turing uses AD is this:

 - wrap the sampler in a SamplingContext
 - wrap the model and the context in a LogDensityFunction
 - when we need to calculate `logdensity_and_gradient`, get the context of the LogDensityFunction, get the sampler of the context, and get the AD type of the sampler
 - construct a `LogDensityProblemsAD.ADgradient` from that and finally calculate the gradient

Specifying the AD type as part of the model will let us simplify this loop and in particular it will enable things like (3) above.

## Problems

The obvious, huge, problem with this is that we would need to rework the whole sampling interface. e.g. instead of

`sample(f(), NUTS(; adtype=AutoMooncake())`

we would need to write something like

`sample(setadtype(f(), AutoMooncake()), NUTS()`.

In an ideal world I would actually prefer this. Of course, this would also be horribly breaking.

The other drawback is that we can't mix and match adtypes, so we can't do this:

`sample(f(), Gibbs(x=NUTS(; adtype=AutoForwardDiff()), y=NUTS(; adtype=AutoMooncake())))`

I don't see a convincing reason why someone would want to do this, though, so I don't feel bad about removing this capability.

penelopeysm:
I have something of a plan to gradually shift us from declaring adtype in samplers to declaring adtype in models. Let's discuss this at a Monday meeting, I think we should probably get it right.

penelopeysm:
I no longer think this is the right way to do it, so closing. But I will add this to the agenda for Monday's meeting.

PR #163, opened 2025-05-20
Improve error message when initial_params / initial_state has wrong length

penelopeysm:
cf. https://github.com/TuringLang/Turing.jl/issues/2557

`check_initial_...` is only used when sampling multiple chains (it doesn't check the actual number of parameters in the model because AbstractMCMC has no way of knowing this), so the old error message was arguably misleading.


sunxd3 (approved):
LG!

PR #174, opened 2025-09-15
Add `chain_number` keyword argument when performing multi-chain sampling

penelopeysm:
Callbacks in general have no knowledge of which chain is being sampled, meaning that there is no way to differentiate callbacks being called from different chains:

https://github.com/TuringLang/AbstractMCMC.jl/blob/33fdad814e999bdccd0b73ebf1195434da6a17c8/src/sample.jl#L235-L237

This PR makes the multiple-chain `sample` method pass a keyword argument, `chain_number`, to the single-chain `sample`. In turn this keyword argument is propagated to the callback so it can be accessed.

IMO the _better_ solution is to actually make it a _positional_ argument to the callback (when sampling a single chain it can either just be `1` or `missing`) but that would be a breaking change and I'm unsure if we want to do that in AbstractMCMC just for this. A keyword argument would just be a minor version bump. Thoughts welcome!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probabilistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "5.7.3"
+version = "5.8.0"
 
 [deps]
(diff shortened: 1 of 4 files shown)


sunxd3 (approved):
I see no issue, make a ton of sense

penelopeysm:
Thanks! I know, it's quite tricky to decide sometimes 😅

PR #182, opened 2025-10-24
Add `AbstractMCMC.getstats` and `AbstractMCMC.requires_unconstrained_space`

penelopeysm:
In general there are two things that samplers return per iteration: parameter values, and sampler statistics.

`getparams` was added previously to cover the first one, but there's no interface function for the second.

Currently, for Turing's external sampler interface, external samplers are expected to implement `Turing.Inference.getparams` and `Turing.Inference.getstats`

This is a mistake, because external samplers shouldn't need to depend on Turing, they should only depend on AbstractMCMC and LogDensityProblems (because they only need to take an `AbstractMCMC.LogDensityModel` and call `LogDensityProblems.logdensity` on its wrapped model). I'm now trying to unravel this, hence the need for this function.

In the same vein, this PR also moves `requires_unconstrained_space` from Turing to here. I'm less certain about this. Please see Slack for discussion. But I think that whether a sampler requires unconstrained space or not is a property of the sampler (and thus belongs in the package that defines AbstractSampler), not a property of how it interacts with Turing.

### DIFF
--- a/src/AbstractMCMC.jl
+++ b/src/AbstractMCMC.jl
@@ -106,7 +106,14 @@ See also [`sample`](@ref).
 struct MCMCSerial <: AbstractMCMCEnsemble end
 
 """
-    getparams([model::AbstractModel, ]state)
+    requires_unconstrained_space(sampler::AbstractSampler)::Bool
+
+Return `true` if the given sampler must run in unconstrained space. Defaults to true.
+"""
+requires_unconstrained_space(::AbstractSampler) = true
+
+"""
+    getparams([model::AbstractModel, ]state)::Vector{<:Real}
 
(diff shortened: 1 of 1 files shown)


src/AbstractMCMC.jl
```diff
@@ -116,6 +123,13 @@ function getparams(model::AbstractModel, state)
     return getparams(state)
 end
 
+"""
+    getstats(state)::NamedTuple
+
+Retrieve sampler statistics from the sampler's `state` as a `NamedTuple`.
+"""
+function getstats end
```
yebai: I'd hope we have more specific guidelines on how samplers should implement or use this API.

PR #186, opened 2025-12-16
add chain name to kwargs of StatsBase.sample

chang-ph:
Add `chain_name` to the kwargs of `StatsBase.sample`.

### The background

We are capturing the live sampling data by implementing a custom `callback` for this method:

```
 [16] sample(
    rng::Random.AbstractRNG, 
    model::AbstractMCMC.LogDensityModel, 
    sampler::AdvancedHMC.AbstractHMCSampler, 
    N::Integer; 
    n_adapts, 
    progress, 
    verbose, 
    callback,  #### our own implementation
    kwargs...)
     @ ~/.julia/packages/AdvancedHMC/sH6aj/src/abstractmcmc.jl:39
```

We want to group sample data by chain. But currently there are no fields in parameters (when calling `callback`) that tells us which chain the data belongs to:

```julia
# https://github.com/TuringLang/AbstractMCMC.jl/blob/main/src/sample.jl#L219
callback === nothing || callback(rng, model, sampler, sample, state, 1; kwargs...)
```

### Changes

1. Users are allowed to provide a `chain_name` argument to the `sample` method. If they don't, a default value (e.g.,`chain#`) is used.
2. When sampling with multiple chains, each chain is assigned an unique index number. The user provided `chain_name` and the index number are combined into a single `chain_name` parameter to the call to `StatsBase.sample`.
3. By doing this, the parameters `callback` will have the `chain_name` in the kwargs. So we can now distinguish the data from different chains.

### DIFF
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -437,6 +437,13 @@ function mcmcsample(
         @warn "Number of chains ($nchains) is greater than number of samples per chain ($N)"
     end
 
+    chain_name_prefix = "chain#"
+    if haskey(kwargs, :chain_name)
+        chain_name_prefix = kwargs[:chain_name]
+        kwargs = Dict(kwargs)
+        delete!(kwargs, :chain_name)
+    end
+
     # Determine default progress bar style.
     if progress == true
         progress = :overall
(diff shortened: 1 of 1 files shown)


chang-ph:
Just noticed https://github.com/TuringLang/AbstractMCMC.jl/pull/174.

chang-ph:
Thanks. It solved my issue perfectly. I didn't notice #174 when I created this PR. Sorry for the inconvenience.

PR #204, opened 2026-05-20
Add AbstractMCMC-based Gibbs combinator

singhharsh1708:
Add model-agnostic Gibbs combinator

Turing.jl and JuliaBUGS.jl currently each have their own separate Gibbs implementation with no shared code. This PR adds a single Gibbs combinator to AbstractMCMC that any PPL can plug into.

The extension interface is three functions per model type:

condition(model, target_vars, global_values) — return a model conditioned on non-target variables
_init_global_values — initialise the global parameter store from the first component's state
_update_global_values — merge new sampled values back into the global store
Component samplers only need the existing getparams/setparams!! interface.

Turing.jl and JuliaBUGS.jl wiring is in separate PRs.

### DIFF
--- a/src/gibbs.jl
+++ b/src/gibbs.jl
@@ -1,4 +1,3 @@
-export Gibbs, GibbsState
 
 """
     AbstractMCMC.condition(model, target_varnames, global_values)
@@ -80,7 +79,9 @@ function step(
     sub_states = _gibbs_initial_steps(
         rng, model, sampler.varnames, sampler.samplers, global_values; kwargs...
     )
-    global_values = _collect_global_values(model, sampler.varnames, sampler.samplers, sub_states)
+    global_values = _collect_global_values(
+        model, sampler.varnames, sampler.samplers, sub_states
+    )
     return _build_gibbs_transition(global_values), GibbsState(global_values, sub_states)
 end
 
@@ -92,13 +93,20 @@ function step(
     kwargs...,
 )
     global_values, sub_states = _gibbs_sweep(
-        rng, model, sampler.varnames, sampler.samplers, state.sub_states, state.global_values;
+        rng,
+        model,
+        sampler.varnames,
+        sampler.samplers,
+        state.sub_states,
+        state.global_values;
         kwargs...,
     )
     return _build_gibbs_transition(global_values), GibbsState(global_values, sub_states)
 end
 
-function _gibbs_sweep(rng, model, varname_groups, samplers, sub_states, global_values; kwargs...)
+function _gibbs_sweep(
+    rng, model, varname_groups, samplers, sub_states, global_values; kwargs...
+)
     new_sub_states = ()
     for i in eachindex(varname_groups)
         target_vars = varname_groups[i]
@@ -110,13 +118,17 @@ function _gibbs_sweep(rng, model, varname_groups, samplers, sub_states, global_v
         _, new_sub_state = step(rng, cond_model, spl, synced_state; kwargs...)
 
         new_params = getparams(cond_model, new_sub_state)
-        global_values = _update_global_values(model, global_values, target_vars, cond_model, new_params)
+        global_values = _update_global_values(
+            model, global_values, target_vars, cond_model, new_params
+        )
         new_sub_states = (new_sub_states..., new_sub_state)
     end
     return global_values, new_sub_states
 end
 
-function _gibbs_initial_steps(rng, model, varname_groups, samplers, global_values; kwargs...)
+function _gibbs_initial_steps(
+    rng, model, varname_groups, samplers, global_values; kwargs...
+)
     sub_states = ()
     for i in eachindex(varname_groups)
         target_vars = varname_groups[i]
@@ -129,7 +141,9 @@ function _gibbs_initial_steps(rng, model, varname_groups, samplers, global_value
             global_values = _init_global_values(model, target_vars, cond_model, sub_state)
         else
             new_params = getparams(cond_model, sub_state)
-            global_values = _update_global_values(model, global_values, target_vars, cond_model, new_params)
+            global_values = _update_global_values(
+                model, global_values, target_vars, cond_model, new_params
+            )
         end
         sub_states = (sub_states..., sub_state)
     end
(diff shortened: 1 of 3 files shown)


test/gibbs_test.jl
shravanngoswamii, line 1: not included in test/runtests.jl, so you only know this PR "works" because of whatever you ran locally.

test/gibbs_test.jl
```diff
@@ -0,0 +1,57 @@
+using Test
+using Random
+using AbstractMCMC
+using Turing
```
shravanngoswamii, line 4: Project.toml's [extras] / test target not updated

src/gibbs.jl
```diff
@@ -0,0 +1,196 @@
...
+
+Return a new model conditioned on all variables *not* in `target_varnames`,
+using the values in `global_values`.
+
+`target_varnames` is a vector of variable identifiers. `global_values` is an
+associative collection mapping identifiers to their current raw values.
+
+Downstream packages implement this per model type. There is no default.
+"""
+function condition end
```
shravanngoswamii, line 13: Are you sure `condition` belongs here? AbstractMCMC's design (https://turinglang.org/AbstractMCMC.jl/stable/design/) is deliberately minimal and PPL-agnostic. `condition` is a PPL concept and `AbstractPPL.condition` already exists.

src/gibbs.jl
```diff
@@ -0,0 +1,196 @@
...
+Component samplers must implement:
+- `AbstractMCMC.step(rng, conditioned_model, sampler[, state])`
+- `AbstractMCMC.getparams(conditioned_model, state) → Vector{<:Real}`
+- `AbstractMCMC.setparams!!(conditioned_model, state, params) → state`
+
+# Example
+
+```julia
+sampler = Gibbs(@varname(μ) => NUTS(), @varname(σ) => MH())
+chain = sample(model, sampler, 1000)
```
shravanngoswamii, line 33: `@varname`, `NUTS`, `MH`, and `model` are not defined in AbstractMCMC. A reader cannot run this example using only this package. Either rewrite it with AbstractMCMC primitives, or move it to a "Downstream usage" section that names Turing explicitly.

src/gibbs.jl
```diff
@@ -0,0 +1,196 @@
...
+```
+"""
+struct Gibbs{N,S} <: AbstractSampler
```
shravanngoswamii, line 36: Constructor validation is too thin. The docstring says "All model variables must be covered by exactly one group," but the only check is `length(varnames) == length(samplers)`. Overlapping groups, empty groups, and zero pairs all silently produce undefined behaviour. Please validate that the groups are disjoint and non-empty.

shravanngoswamii (commented):
Being candid since this targets AbstractMCMC, one of the most depended-on packages in the Turing ecosystem -- a single bad API addition here affects many downstream packages. This PR has both design problems and real correctness/quality bugs. Please settle the interface design first, and fix the bugs listed below before the next review round.

- Nothing is exported. If this is meant as a public API, export it. If it is not, the docstring should not claim it is.
- `condition` will collide with `AbstractPPL.condition` for any user doing `using AbstractMCMC, AbstractPPL`.
- `_init_global_values` / `_update_global_values` / `_build_gibbs_transition` are part of the extension interface that downstream packages must implement, but the leading underscore signals "private." Drop the underscore, or move them to an internal module.
- `initial_params` is reused for `global_values`. Everywhere else in AbstractMCMC, `initial_params` is a flat `Vector{<:Real}`. Pick a new kwarg name.
- The docstring says `global_values` is "an associative collection mapping identifiers to their current raw values," and then on the first init step you pass `nothing`. Either fix the doc or fix the call site.
- `GibbsState` does not itself implement `getparams` / `setparams!!` / `getstats`. That means it cannot be nested inside another `Gibbs`, cannot be used by callbacks that expect `getparams(state)`, and breaks the same contract this PR demands of its components.
- `using Turing` in tests cannot work: Turing depends on AbstractMCMC, so this is a hard circular dependency. The tests must be written using AbstractMCMC primitives only, plus a small hand-written model and a small sampler that implements the contract.

shravanngoswamii:
@singhharsh1708 Please read [CONTRIBUTING.md](https://github.com/TuringLang/AbstractMCMC.jl?tab=contributing-ov-file#contributing-to-turingjl-organization-repositories) and organization wide `IMPORTANT` section in organization README: https://github.com/TuringLang

It's fine if you use AI for this but it was a GSoC project meant to learn how to engage in open source softwares and how to understand them and contribute positively. Atleast discuss this with team before implementing.

shravanngoswamii:
@sunxd3 Can you see if this should be continued here or discussed first?

singhharsh1708:
@shravanngoswamii You're right — I should have opened a discussion issue first before sending code. I used AI assistance to help implement this, but I understand that for a package this central, the interface design needs to be agreed on with the team before any implementation. I'll close these PRs and open a design discussion first. will that be fine?

shravanngoswamii:
> I'll close these PRs and open a design discussion first. will that be fine?

Yep, @sunxd3 Can you please see how this should be continued...

sunxd3:
A constraint is that unless we can find an interface that DynamicPPL can use without (1) performance penalty and (2) support all the inference algorithms (tricky ones are the particle/SMC), it won't get adopted by Turing.

This is not trivial, otherwise Penny (Markus and Tor) would have done it already.

Start by designing is definitely the right approach and one of the first questions to think about is why we need the GibbsContext in Turing.

shravanngoswamii:
@singhharsh1708 Your approach is correct, you just need to have to some more context of how these are shared and should be designed across the ecosystem. I will not close it. Let's discuss this first in a issue. Feel free to open and wait for response from the team.

yebai:
This is a challenging technical task and should be implemented under the close supervision of a core team member. Please reach out via email if you are interested.

singhharsh1708:
> This is a challenging technical task and should be implemented under the close supervision of a core team member. Please reach out via email if you are interested.

yes i am interested in this project and would like to work on it.

PR #212, opened 2026-07-13
Allow structured parameter containers in ParamsWithStats

sunxd3:
`ParamsWithStats` required `params` to be a `NamedTuple`, so parameter containers with non-`Symbol` keys could not be stored without lossy conversion. This PR relaxes the struct to `ParamsWithStats{P,S<:NamedTuple,E<:NamedTuple}`: the parameter container can be any type implementing `pairs` and `isempty`, with keys that have a meaningful string form so name-based filtering and logging work. It also adds a two-argument constructor `ParamsWithStats(params, stats)` (empty `extras`) and documents that `pairs(pws)` may yield mixed key types, so consumers should iterate generically.

This is intended to be non-breaking: existing constructors and the `Vector{<:Real}`/`Vector{<:Pair}` normalizations are unchanged, and equality keeps the default struct behavior. Bumps the version to 5.16.0.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 keywords = ["markov chain monte carlo", "probabilistic programming"]
 license = "MIT"
 desc = "A lightweight interface for common MCMC methods."
-version = "5.15.1"
+version = "5.16.0"
 
 [deps]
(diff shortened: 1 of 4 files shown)


yebai (commented):
Thanks, @sunxd3 - good work!

sunxd3:
This should allow us to start trying to use `AbstractMCMC.ParamsWithStats` in DynamicPPL (currently the two `ParamsWithStats` are different type).

PR #215, opened 2026-07-22
Isolate initial_params/initial_state per chain in multi-chain sampling

### LINKED ISSUE #214: Multi-chain `sample` shares one `initial_params` object across chains — segfaults under ForwardDiff
Multichain `mcmcsample` deep-copies `model` and `sampler` per chain, but passes each `initial_params[i]` through uncopied. So `fill(init, nchains)` — a natural idiom — hands the *same* object to every chain. With Turing + ForwardDiff this reliably segfaults (signal 11) within a few iterations; a distinct object per chain is clean.

MWE (Julia 1.12):

```julia
using Turing, StableRNGs
const TU = Turing.DynamicPPL.TestUtils

for rep in 1:12, model in TU.DEMO_MODELS
    pm = TU.posterior_mean(model)
    sample(StableRNG(42), model, NUTS(), MCMCSerial(), 2000, 4;
           initial_params = fill(InitFromParams(pm), 4),   # ONE object shared across chains
           discard_initial = 1000, thinning = 10, progress = false)
end
```

```
signal (11): Segmentation fault
gc_mark_obj8 -> gc_mark_outrefs -> _jl_gc_collect   (in DynamicPPL's RawValueAccumulator alloc)
```

Clean under any one of: `initial_params = [InitFromParams(pm) for _ in 1:4]` (distinct per chain); a single chain; or ReverseDiff instead of ForwardDiff.

**This is almost certainly an upstream Julia GC bug** — heap corruption surfaced during GC mark, only under ForwardDiff, non-deterministic, and independent of the shared object's *values* (the `fill` and comprehension vectors alias the same, never-mutated arrays; only the object-graph shape differs). AbstractMCMC can't fix that. But it can **work around it** by isolating `initial_params[i]` (and `initial_state[i]`) per chain via `deepcopy`, consistent with the existing per-chain isolation of `model`/`sampler` — which removes this footgun regardless of the Julia fix.

Versions: AbstractMCMC 5.15.1, DynamicPPL 0.42.2, ForwardDiff 1.4.1, Julia 1.12.6 (aarch64-darwin).

yebai:
Multi-chain `mcmcsample` already deep-copies `model` and `sampler` per chain, but passed each `initial_params[i]`/`initial_state[i]` through uncopied. So a shared init object — e.g. `fill(init, nchains)`, a natural idiom — reaches every chain. With Turing + ForwardDiff this reliably segfaults (heap corruption surfaced during GC mark) within a few iterations; a distinct object per chain is clean.

This `deepcopy`s the per-chain init so each chain gets its own copy, matching the existing per-chain isolation of `model`/`sampler`. The underlying corruption is almost certainly an upstream Julia GC bug — it only occurs under ForwardDiff, is non-deterministic, and is independent of the init's values (the shared arrays are never mutated) — so this can only work around it, but per-chain isolation removes the footgun regardless of the Julia fix.

It's hard to add tests for these as it is a Julia GC bug. 

Fixes #214.

### DIFF
--- a/src/sample.jl
+++ b/src/sample.jl
@@ -666,7 +666,9 @@ function mcmcsample(
                                 )
                             end
 
-                            # Sample a chain and save it to the vector.
+                            # Sample a chain and save it to the vector. `deepcopy` the init
+                            # per chain -- a shared init object segfaults under ForwardDiff
+                            # (Julia GC bug; workaround for #214).
                             chains[chainidx] = StatsBase.sample(
                                 _rng,
                                 _model,
@@ -676,12 +678,12 @@ function mcmcsample(
                                 initial_params=if initial_params === nothing
                                     nothing
                                 else
-                                    initial_params[chainidx]
+                                    deepcopy(initial_params[chainidx])
                                 end,
                                 initial_state=if initial_state === nothing
                                     nothing
                                 else
-                                    initial_state[chainidx]
+                                    deepcopy(initial_state[chainidx])
                                 end,
                                 chain_number=chainidx,
                                 kwargs...,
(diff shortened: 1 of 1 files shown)


yebai:
Closed in favour of https://github.com/TuringLang/DynamicPPL.jl/pull/1440

